PR #17, opened 2015-02-20
Add debian:jessie support for 1.7.10

nickjj:
A lot of official programming language run-times such as
Golang, Ruby and NodeJS use Jessie as a base rather than
Wheezy.

To ensure the footprint remains as small as it can be this
patch allows you to optionally pick Jessie.


thresheek:
I'd rather not install package built for wheezy on a jessie machine.  We're in a process of establishing an infrastructure to do proper jessie-based builds and once jessie hits stable, I will switch nginx-docker image to use it.

nickjj:
Fair enough, sounds good.

PR #20, opened 2015-04-16
Use http for PGP key download

albers:
The hkp service (port 11371) is likely to be blocked by firewalls.
To enable building of this image in a corporate environment,
http should be used instead.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM debian:wheezy
 
 MAINTAINER NGINX Docker Maintainers "docker-maint@nginx.com"
 
-RUN apt-key adv --keyserver pgp.mit.edu --recv-keys 573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62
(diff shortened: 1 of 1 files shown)


thresheek:
Thank you, that makes sense.

sherter:
The question here really is: why do you want to build this image by yourself?

hacfi:
@sherter To adapt it to your needs...I’m using another base image for example.

sherter:
@hacfi But then you need you own Dockerfile either way. So why "pollute" this Dockerfile?

albers:
I see this repo as the primary source of information of how to build an image with an up-to-date Nginx. It's kind of the reference implementation of an nginx image.
Of course it is not your responsibility to ensure that this image can be built outside its official build infrastructure. But it would be _nice_ if it could.

There are valid usecases that require customization of the Dockerfile (like the one @sherter mentioned). More important, people will copy the installation instructions. If one of the build steps unnecissarily uses a firewall-unfriendly protocol, this will cause trouble for those people. I just wanted to contribute back my solution for the problems I experienced in this context. Feel free to revert if you think this change is a regression.

Issue #3, opened 2014-09-25
Can you include 1.6 branch as well


yankeeinlondon:
I will likely be moving to 1.7.x but for now I'm still entrenched in 1.6 and was hoping you'd support a 1.6 branch.

PR #21, opened 2015-04-22
Bump to 1.8.0 (stable)

ffung:


thresheek:
I will bump to 1.9.0 (mainline) when it's out.

ffung:
ok, you don't want to publish a stable version in the hub?

btw there was also a fix in this PR for suppressing the TERM warnings during build

thresheek:
Yes, the idea is to use mainline here, see https://github.com/nginxinc/docker-nginx/issues/3.

as for TERM I don't really want to introduce another layer for a cosmetic change.

ffung:
ok fair enough ;-)

PR #37, opened 2015-08-20
Add AUTOINDEX=ON environment variable to index static files

mildred:
Added an entrypoint that updates the configuration depending on environment variables.

If the environment variable `AUTOINDEX` is set to `ON` this will generate a `autoindex on;` on the default configuration.


thresheek:
I would rather keep main nginx docker image free of various task-specific hacks.  Please use FROM and apply any changes you might need in your environment locally.

mildred:
Ok, that was in fact my original thought but when I was a comment on DockerHub telling us to contribute instead of creating new versions of containers, I thought I would try a pull request.

PR #53, opened 2015-12-10
Update to Nginx 1.9.9

md5:


md5:
Looks like this got done in https://github.com/nginxinc/docker-nginx/commit/a8b6da8425c4a41a5dedb1fb52e429232a55ad41

PR #63, opened 2016-01-26
Add alpine variant

ncopa:
Try stay as close as possible to the debian variant, so we also include
the nginx-debug binary in the same container.


ncopa:
i suppose we can close this?

PR #72, opened 2016-03-09
Fix "gpg" usage to stop relying on deprecated and insecure behavior

tianon:
See also docker-library/official-images#1420.

I didn't think this was urgent enough to force you to round-trip on it again for getting https://github.com/docker-library/official-images/pull/1462 through, but wanted to get it on your radar for the next update to the image. :smile:

I also swapped `alpine` to use `:3.3` explicitly so that we don't run into any unexpected problems (for example, if 3.4 were to come out with a new version of gcc and NGINX fails to compile for whatever reason -- just to be on the safer side). :+1:

### DIFF
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:3.3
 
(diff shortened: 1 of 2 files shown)


thresheek:
Merged, thanks!

PR #103, opened 2016-07-07
Use all cores when compiling

mattrobenolt:
:rocket:

### DIFF
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -77,14 +77,14 @@ RUN \
 	&& rm nginx.tar.gz \
 	&& cd /usr/src/nginx-$NGINX_VERSION \
 	&& ./configure $CONFIG --with-debug \
-	&& make \
+	&& make -j$(getconf _NPROCESSORS_ONLN) \
 	&& mv objs/nginx objs/nginx-debug \
 	&& mv objs/ngx_http_xslt_filter_module.so objs/ngx_http_xslt_filter_module-debug.so \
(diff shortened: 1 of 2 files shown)


thresheek:
Sorry, I don't think this is a good idea.  While it does speed up things, nginx compilation time is not long anyway.

Also, to be a good citizen on auto-builders / docker official images builders, I would rather not abuse them this way.  I've also checked a couple of other projects that have official images and are built, they don't use parallel jobs.

mattrobenolt:
For reference from the official library: https://github.com/docker-library/memcached/blob/master/alpine/Dockerfile#L25

:)

mattrobenolt:
And python: https://github.com/docker-library/python/blob/master/3.5/alpine/Dockerfile#L44 probably others if I dug through them all.

thresheek:
Heh, interesting.  I've checked haproxy and couchdb and they both do not have it set.  I'll check with the library folk.

mattrobenolt:
Paging @tianon !

tianon:
We do use `-j$(nproc)`/`-j$(getconf _NPROCESSORS_ONLN)` pretty commonly; anywhere we don't within "docker-library" repos is likely just a case of it compiling so fast that `-j` doesn't make much difference anyhow (`haproxy` and `redis` come to mind).

luxas:
Any chance this could be merged?

thresheek:
Fixed by https://github.com/nginxinc/docker-nginx/commit/e117bd83e9befe5582bc1da8f72248398fffa16c

Thanks!

PR #105, opened 2016-07-07
Extract only envsubst out of gettext

mattrobenolt:
This results in a reduction of ~5.5MB from the image.

```
Before: 59688904 bytes
After:  54104509 bytes
```

### DIFF
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -99,17 +99,26 @@ RUN \
 	&& ln -s ../../usr/lib/nginx/modules /etc/nginx/modules \
 	&& strip /usr/sbin/nginx* \
 	&& strip /usr/lib/nginx/modules/*.so \
+	&& rm -rf /usr/src/nginx-$NGINX_VERSION \
+	\
+	# Bring in gettext so we can get `envsubst`, then throw
+	# the rest away. To do this, we need to install `gettext`
+	# then move `envsubst` out of the way so `gettext` can
+	# be deleted completely, then move `envsubst` back.
+	&& apk add --no-cache --virtual .gettext gettext \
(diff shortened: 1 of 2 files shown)


thresheek:
Thank you!

I've pushed the patch manually.

md5:
This seems like it could be problematic if someone needs to have the rest of `gettext` in a derived image. I wonder if it would be possible to split `envsubst` into a separate package in Alpine itself.

md5:
I also wonder how much of the size savings are actually coming from deleting the Nginx source and adding `--no-cache` to the runDeps installation.

thresheek:
Well this is a bit out of scope for the current image; would alpine package manager overwrite the provided /bin/envsubst if gettext is installed by someone using FROM with this image?

thresheek:
nginx source was deleted in the prior versions as well, so it wont make a difference.  the diff is kinda confusing thou.

md5:
I believe the -f flag would be needed, which would only be discoverable
through reading the Dockerfile and seeing that the package is only
partially installed or through trial and error.

md5:
@clandmeter What do you think of breaking out `envsubst` into a `getttext-envsubst` package? I see that there are already a number of subpackages of `gettext`, including `gettext-asprintf`.

md5:
@thresheek I see what you mean about `/usr/src/nginx-$NGINX_VERSION` in the diff. I got confused reading it on my phone.

mattrobenolt:
> What do you think of breaking out envsubst into a getttext-envsubst package?

Also, I'd be open to just moving `envsubst` into say, `/usr/local/bin` instead of putting it back into `/usr/bin`, that we re-installing `gettext` won't have to overwrite anything.

Is that a fair compromise?

PR #106, opened 2016-07-07
Move CONFIG/GPG_KEYS out of image ENV vars

mattrobenolt:
There's no reason for these to exist as environment variables, and
CONFIG is pretty clunky and useless outside of build time.

I'd argue that if someone wanted to be able to inspect this information,
it could be retrieved from `nginx -V`.

Another point is that these don't exist for the `debian:jessie` based
images either.

### DIFF
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -4,53 +4,51 @@ MAINTAINER NGINX Docker Maintainers "docker-maint@nginx.com"
 
 ENV NGINX_VERSION 1.11.1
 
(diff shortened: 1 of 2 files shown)


thresheek:
Thank you!

I've pushed the patch manually.

cameronmalek:
Why was `NGINX_VERSION` excluded from this? I can't think of a good reason.

mattrobenolt:
It's much more common practice to leave the software version in environment variables. I don't know the reasoning tbh, but it's exhibited in every official image, so I choose to follow convention.

cameronmalek:
Okay, thanks.

PR #111, opened 2016-07-22
add parallel process make with "-j" param

jaschweder:
For better performance when building can add the "-j" param using the "(number of processors) + 1" formula


jaschweder:
duplicated #103

Issue #76, opened 2016-03-29
ALPN support for HTTP/2


mxmo0rhuhn:
Would it be possible to include a OpenSSL version >= 1.0.2 to make ALPN work?
Due to the [official documentation](http://nginx.org/en/docs/http/ngx_http_v2_module.html) it's needed for HTTP/2 over TLS: 

> Note that accepting HTTP/2 connections over TLS requires the “Application-Layer Protocol Negotiation” (ALPN) TLS extension support, which is available only since OpenSSL version 1.0.2. 

The current version in the **nginx:1.9.12** image is: 

```
$ openssl version
OpenSSL 1.0.1k 8 Jan 2015
```

thresheek:
Note that HTTP/2 right now works via NPN, which does not require openssl 1.0.2 and most browsers still support it at the time of writing.

The support for ALPN will arrive in nginx:latest as soon as Debian Stretch with openssl 1.0.2 is released.
If you really need ALPN with nginx official, please use the alpine variant, as it has openssl 1.0.2.

mxmo0rhuhn:
Thanks a lot for the fast reply! 
I will have a look at the alpine variant.

hermansje:
Starting May 15th, Chrome will drop support for SPDY and NPN [1].
Does this mean everyone wanting something better than HTTP/1.1 using this image should switch to the Alpine version?

[1] https://blog.chromium.org/2016/02/transitioning-from-spdy-to-http2.html

thresheek:
Not really, you can still use HTTPS.

teohhanhui:
I agree with this: https://plus.google.com/+AmbroosVaes/posts/TLJJ41Ghg9R

There does't seem to be any workaround for now except building OpenSSL ourselves (is it easy?) or switching to the Alpine variant (which is alien to most people).

hermansje:
I tried the Alpine version and HTTP/2 and ALPN are working. I only had to change the `FROM` instruction to reference an Alpine version.

thresheek:
@teohhanhui indeed. There are a lots of people with the same arguments like you linked, but apparently Google does not care at all (not that is surprising).

One of the workarounds for nginx could be using Ubuntu 16.04 as a base image, but I'm not really sure how much breakage that will cause - I would expect many people actually do FROM nginx:latest and use debian-specific things there. I'm also sure HTTP2 does not justify that breakage - after all, browsers will still show your website.

fruitl00p:
[...]

PR #129, opened 2016-10-30
Bring http2 support back to chrome.

JamesKyburz:
How about adding an official image for ubuntu/xenial?

Then we could get http2 support in chrome without building our own custom docker images?

This PR is in response to https://github.com/nginxinc/docker-nginx/issues/76


thresheek:
Please use alpine tags.

I plan to move to the newer openssl-based image when Debian 9 is released (around March 2017 it seems).

JamesKyburz:
Ok thanks.

PR #136, opened 2016-12-28
Upgrade alpine to v3.5

PeterDaveHello:


thresheek:
Is there a real problem solved by this upgrade?

PeterDaveHello:
@thresheek nope, but Alpine v3.5 is even smaller than v3.4, could be a good reason! Since there are so many people pulling the images every day!

PeterDaveHello:
@thresheek is it good to go? Thanks.

justsml:
Solves a real problem: the openssl-dev version in 3.4 doesn't support ALPN for full HTTP2 compliance.

Alpine v3.5 advances available cryptography enough to justify quick upgrades IMHO

yosifkit:
If there are many users that would install additional packages `FROM nginx`, then I would think it might be better to wait for a major bump to nginx. See also: https://github.com/docker-library/golang/pull/131#issuecomment-270677556

ThijsBorst:
certbot 0.9.x will be supported as well, which has support for nginx... Really want this :-)

thresheek:
Done starting with nginx:1.13.0 / mainline and nginx:1.12.0 / stable.

PR #142, opened 2017-02-06
Use Entrypoint Instead of Cmd for Nginx Container

goettl79:
After this change I'm be able to start nginx like this:

docker run nginx:1.10 -c /mypath/nginx.conf

Currently i'd need to do:

docker run nginx:1.10 nginx -c /mypath/nginx.conf

### DIFF
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -132,4 +132,4 @@ COPY nginx.vh.default.conf /etc/nginx/conf.d/default.conf
 
 EXPOSE 80 443
 
-CMD ["nginx", "-g", "daemon off;"]
+ENTRYPOINT ["nginx", "-g", "daemon off;"]
--- a/mainline/jessie/Dockerfile
+++ b/mainline/jessie/Dockerfile
@@ -24,4 +24,4 @@ RUN ln -sf /dev/stdout /var/log/nginx/access.log \
 
 EXPOSE 80 443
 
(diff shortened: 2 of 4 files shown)


tianon:
Be sure to keep https://github.com/docker-library/official-images/blob/1eb98a3332bcaa7d23338ee7c05d15f882b53acd/README.md#consistency in mind while reviewing this. :heart:

goettl79:
I think point 3 of the list in https://github.com/docker-library/official-images/blob/1eb98a3332bcaa7d23338ee7c05d15f882b53acd/README.md#consistency is the nginx scenario. ( ... the image only contains the main executable and its linked libraries (ie no shell) ... )  Should be valid to use Entrypoint then.

yosifkit:
No, point 3 doesn't apply unless there is no shell; these images are from Debian and Alpine, so there is definitely a shell available.  It usually only applies to images `FROM scratch`.

goettl79:
Well,then this pull request is against the docker consistency rules for official images. `docker run nginx:1.10 bash` won't work.
From a developer / non administrator perspective this rule is however questionable. Won't need a shell first to feel comfortable. `docker run nginx:1.10` is to my taste much better understandable as `docker run nginx:1.10 nginx`, to put my two cents in. But this might be the wrong place to discuss this. Shall I close the pull request?

thresheek:
I think it's more natural to have it the way it currently works.

PR #155, opened 2017-03-16
add Dockerfile for building windows image

friism:
Opening this for discussion.

cc @tianon @stefanScherer

### DIFF
--- a/stable/windows/Dockerfile
+++ b/stable/windows/Dockerfile
@@ -1,0 +1,16 @@
+FROM microsoft/windowsservercore
+
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
+ENV NGINX_VERSION 1.10.3
+RUN Invoke-WebRequest -Uri https://nginx.org/download/nginx-$Env:NGINX_VERSION.zip -OutFile nginx.zip ; \
+	Expand-Archive nginx.zip -DestinationPath $Env:ProgramFiles ; \
+	Remove-Item -Force nginx.zip ; \
+	Move-Item $Env:ProgramFiles\nginx-* $Env:ProgramFiles\nginx
+
+RUN setx /M PATH $($Env:PATH + ';' + $Env:ProgramFiles + '\nginx')
+
+EXPOSE 80 443
+WORKDIR "C:\Program Files\nginx"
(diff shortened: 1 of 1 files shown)


stable/windows/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+	Expand-Archive nginx.zip -DestinationPath $Env:ProgramFiles ; \
+	Remove-Item -Force nginx.zip ; \
+	Move-Item $Env:ProgramFiles\nginx-* $Env:ProgramFiles\nginx
+
+RUN setx /M PATH $($Env:PATH + ';' + $Env:ProgramFiles + '\nginx')
+
+EXPOSE 80 443
+WORKDIR "C:\Program Files\nginx"
+CMD Start-Process -NoNewWindow -FilePath nginx.exe ; \
+	Add-Content logs\access.log 'nginx started...' ; Get-Content -Wait logs\access.log
```
tianon: This is basically `tail -f` -- can't we run `nginx` the same way in Windows that we do in Linux to get it to stay in the foreground and log to stdout/stderr?

friism: @tianon I googled hard, but I don't think there's an elegant way to symlink the logfiles to stdout/err devices on Windows (they just don't exist).

I agree that this solution is super lame. The only reason it would be acceptable is because nginx on Windows is experimental in the first place.

tianon: :sob:

stable/windows/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM microsoft/windowsservercore
```
StefanScherer: Can we run nginx.exe in nanoserver?

friism: @StefanScherer It doesn't work currently, I tried. I'm guessing it uses 32bit APIs and relies on WoW. Someone should try to compile with 64-bit only toolchain for nanoserver.

StefanScherer: ok, would be great for future versions of nginx.

StefanScherer:
I just had a look at http://nginx.org/en/docs/windows.html

> To install nginx/Windows, download the latest mainline version distribution (1.11.10), since the mainline branch of nginx contains all known fixes.

thresheek:
Given that the facts outlined in http://nginx.org/en/docs/windows.html#known_issues still stand, I don't really get the idea of publishing it on docker.

And yes, mainline version should be published, not the stable one.

PR #161, opened 2017-04-20
change uid/gid from nginx to www-data(82)

Amakata:
82 is the standard uid/gid for "www-data" in Alpine
http://git.alpinelinux.org/cgit/aports/tree/main/apache2/apache2.pre-install?h=v3.3.2
http://git.alpinelinux.org/cgit/aports/tree/main/lighttpd/lighttpd.pre-install?h=v3.3.2
http://git.alpinelinux.org/cgit/aports/tree/main/nginx-initscripts/nginx-initscripts.pre-install?h=v3.3.2


thresheek:
https://git.alpinelinux.org/cgit/aports/tree/main/nginx uses nginx as user/group and init scripts you linked (nginx-initscripts) also create nginx user and group.  Why change that?

thresheek:
Feedback timeout.

PR #176, opened 2017-06-09
Use alpine:3.6 base image

marians:
The alpine:3.5 base image has several known vulnerabilities in zlib, with "critical" or "major" severity.

See

- https://hub.docker.com/r/library/alpine/tags/3.5/
- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9841
- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9843
- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9842
- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9840

alpine:3.6 uses a newer zlib version and has no known vulnerabilities. See

https://hub.docker.com/r/library/alpine/tags/3.6/

### DIFF
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.5
+FROM alpine:3.6
 
 MAINTAINER NGINX Docker Maintainers "docker-maint@nginx.com"
 


yosifkit:
Having CVE "vulnerabilities" does not mean it is actually usefully vulnerable. On all 4 cves you link both RedHat Security and Debian Security have marked it as a "minor issue". RedHat marked `zlib` as "will not fix" on RHEL 5, 6, and 7; while Debian has at least fixed stretch/testing (which is still a pre-release anyway).  Alpine 3.6 was released in the last couple weeks so it could apply the fix before a snapshot release.

- https://access.redhat.com/security/cve/CVE-2016-9841
- https://access.redhat.com/security/cve/CVE-2016-9843
- https://access.redhat.com/security/cve/CVE-2016-9842
- https://access.redhat.com/security/cve/CVE-2016-9840
- https://security-tracker.debian.org/tracker/CVE-2016-9841
- https://security-tracker.debian.org/tracker/CVE-2016-9843
- https://security-tracker.debian.org/tracker/CVE-2016-9842
- https://security-tracker.debian.org/tracker/CVE-2016-9840

That said, I think the Nginx maintainer likes to only do major changes (like an OS bump) when there is a version bump of Nginx as well.

marians:
Thanks for the insight! It appears to me as if the bump from 3.4 to 3.5 has also been made without an nginx version change. See https://github.com/nginxinc/docker-nginx/pull/136

thresheek:
#136 was done on nginx version change - from 1.10/1.11 to 1.12/1.13.. That being said, are those CVEs really there in 3.5?  Maybe they're patched out?

romdim:
Hello there! Is there a reason why you always wait for the next version of nginx/ruby/... in order to upgrade the os version as well? Because I don't get why it should be like that.

morganchristiansson:
All nginx docker images are showing up as vulnerable in the docker security scanner

https://hub.docker.com/r/library/nginx/tags/

zyrill:
If you don't want to wait for them to update the official images, here's the Dockerfile code changes required for moving to alpine v3.6: https://github.com/zyrill/docker-nginx/commit/45860be26a33b7a69130f766db8de34ae85a8a8a
Going to v1.13.4 on nginx is trivial, too: https://github.com/zyrill/docker-nginx/commit/11583294b5e5a5fafe100fd214e2b0eff03dc681

thresheek:
AFAICT there are no zlib vulnerabilites at the moment present in mainline:alpine images: https://hub.docker.com/r/library/nginx/tags/1.13.3-alpine/
Hopefully there will be even less when 1.13.4 is pushed - which is why I don't see a reason to jump to 3.6 right now.

PR #187, opened 2017-08-09
Update mainline NGINX to 1.13.4

alessfg:


thresheek:
I think njs was also updated to 0.12, so some dockerfiles will fail.

alessfg:
Bumped NJS to 0.12. Dockerfiles are working now.

PR #188, opened 2017-08-12
Add support for OpenShift compatible NGINX Docker images

alessfg:
Initial implementation of Dockerfiles to produce OpenShift compatible NGINX Docker images

### DIFF
--- a/mainline/openshift/Dockerfile
+++ b/mainline/openshift/Dockerfile
@@ -1,0 +1,32 @@
+FROM docker.io/centos:7
+
+ENV NGINX_VERSION 1.13.4-1.el7
+
+LABEL maintainer="NGINX Docker Maintainers <docker-maint@nginx.com>"
+
+ADD nginx.repo /etc/yum.repos.d/nginx.repo
+
+RUN curl -sO http://nginx.org/keys/nginx_signing.key && \
+    rpm --import ./nginx_signing.key && \
+    yum -y install --setopt=tsflags=nodocs nginx-${NGINX_VERSION}.ngx && \
+    rm -f ./nginx_signing.key && \
+    yum clean all
+
+# forward request and error logs to docker log collector
+RUN ln -sf /dev/stdout /var/log/nginx/access.log && \
+    ln -sf /dev/stderr /var/log/nginx/error.log && \
+# Change pid file location & remove nginx user & change port to 8080
+    sed -i 's/\/var\/run\/nginx.pid/\/var\/cache\/nginx\/nginx.pid/g' /etc/nginx/nginx.conf && \
+    sed -i -e '/user/!b' -e '/nginx/!b' -e '/nginx/d' /etc/nginx/nginx.conf && \
+    sed -i -e '/listen/!b' -e '/80;/!b' -e 's/80;/8080;/' /etc/nginx/conf.d/default.conf && \
+# modify perms for non-root runtime
+    chown -R 998 /var/cache/nginx /etc/nginx && \
+    chmod -R g=u /var/cache/nginx /etc/nginx
+
+VOLUME ["/var/cache/nginx"]
+
+EXPOSE 8080 8443
+
+USER 998
+
+CMD ["nginx", "-g", "daemon off;"]
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -96,4 +96,18 @@ for version in "${versions[@]}"; do
 			Directory: $version/$variant
 		EOE
 	done
+
+	for variant in openshift; do
+		commit="$(dirCommit "$version/$variant")"
+
+		variantAliases=( "${versionAliases[@]/%/-$variant}" )
+		variantAliases=( "${variantAliases[@]//latest-/}" )
+
+		echo
+		cat <<-EOE
+			Tags: $(join ', ' "${variantAliases[@]}")
+			GitCommit: $commit
+			Directory: $version/$variant
+		EOE
+	done
 done
--- a/mainline/openshift/nginx.repo
+++ b/mainline/openshift/nginx.repo
@@ -1,0 +1,5 @@
+[nginx]
+name=nginx repo
+baseurl=http://nginx.org/packages/mainline/centos/7/$basearch/
+gpgcheck=1
+enabled=1
--- a/stable/openshift/Dockerfile
+++ b/stable/openshift/Dockerfile
@@ -1,0 +1,32 @@
+FROM docker.io/centos:7
+
+ENV NGINX_VERSION 1.12.1-1.el7
+
+LABEL maintainer="NGINX Docker Maintainers <docker-maint@nginx.com>"
+
+ADD nginx.repo /etc/yum.repos.d/nginx.repo
+
+RUN curl -sO http://nginx.org/keys/nginx_signing.key && \
+    rpm --import ./nginx_signing.key && \
+    yum -y install --setopt=tsflags=nodocs nginx-${NGINX_VERSION}.ngx && \
+    rm -f ./nginx_signing.key && \
+    yum clean all
+
+# forward request and error logs to docker log collector
+RUN ln -sf /dev/stdout /var/log/nginx/access.log && \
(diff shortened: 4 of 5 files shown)


alessfg:
- [x]  Create NGINX OpenShift Dockerfiles

- [ ]  Update generate-stackbrew-library.sh

tianon:
What's different about OpenShift which makes the existing images incompatible?

alessfg:
OpenShift (or more specifically the wrappers that OpenShift creates around Docker images) does not allow containers that run as root which results in the NGINX images failing at startup. These Dockerfiles have the necessary modifications to run NGINX as non-root.

mainline/openshift/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    sed -i -e '/listen/!b' -e '/80;/!b' -e 's/80;/8080;/' /etc/nginx/conf.d/default.conf && \
+# modify perms for non-root runtime
+    chown -R 998 /var/cache/nginx /etc/nginx && \
+    chmod -R g=u /var/cache/nginx /etc/nginx
+
+VOLUME ["/var/cache/nginx"]
+
+EXPOSE 8080 8443
+
+USER 998
```
thresheek: Where does the magic number 998 come from?  Is it guaranteed to be the same always? Why not to use "nginx" just like we add in rpm post-scripts?

mainline/openshift/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM docker.io/centos:7
+
+ENV NGINX_VERSION 1.13.4-1.el7
+
+LABEL maintainer="NGINX Docker Maintainers <docker-maint@nginx.com>"
+
+ADD nginx.repo /etc/yum.repos.d/nginx.repo
+
+RUN curl -sO http://nginx.org/keys/nginx_signing.key && \
```
thresheek: Insecure.  Please use constructions similar to https://github.com/nginxinc/docker-nginx/blob/master/mainline/stretch/Dockerfile#L11

mainline/openshift/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# forward request and error logs to docker log collector
+RUN ln -sf /dev/stdout /var/log/nginx/access.log && \
+    ln -sf /dev/stderr /var/log/nginx/error.log && \
+# Change pid file location & remove nginx user & change port to 8080
+    sed -i 's/\/var\/run\/nginx.pid/\/var\/cache\/nginx\/nginx.pid/g' /etc/nginx/nginx.conf && \
+    sed -i -e '/user/!b' -e '/nginx/!b' -e '/nginx/d' /etc/nginx/nginx.conf && \
+    sed -i -e '/listen/!b' -e '/80;/!b' -e 's/80;/8080;/' /etc/nginx/conf.d/default.conf && \
+# modify perms for non-root runtime
+    chown -R 998 /var/cache/nginx /etc/nginx && \
```
thresheek: No, that's bad - the files in /etc/nginx will be writable by nginx user, which means in case of vulnerability attacker could rewrite the configurations.

mainline/openshift/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    chmod -R g=u /var/cache/nginx /etc/nginx
```
thresheek: Why?

mainline/openshift/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    rm -f ./nginx_signing.key && \
+    yum clean all
+
+# forward request and error logs to docker log collector
+RUN ln -sf /dev/stdout /var/log/nginx/access.log && \
+    ln -sf /dev/stderr /var/log/nginx/error.log && \
+# Change pid file location & remove nginx user & change port to 8080
+    sed -i 's/\/var\/run\/nginx.pid/\/var\/cache\/nginx\/nginx.pid/g' /etc/nginx/nginx.conf && \
+    sed -i -e '/user/!b' -e '/nginx/!b' -e '/nginx/d' /etc/nginx/nginx.conf && \
+    sed -i -e '/listen/!b' -e '/80;/!b' -e 's/80;/8080;/' /etc/nginx/conf.d/default.conf && \
```
thresheek: This change should be documented as it's a big behaviour change from usual images.

yosifkit:
For the official-images we don't really want images to create extra tags that are only meant for a specific vendor/cloud/ci platform.  We support being able to run an image as non-root and many service images support running as any arbitrary user, like [postgres](https://github.com/docker-library/docs/tree/2809012f34a8141b840f0ad899f591dac81b53f8/postgres#arbitrary---user-notes), but we resist changes that are meant only for a specific environment like https://github.com/docker-library/docker/pull/12.

If the goal is to make it run as non-root, I was able to get it to run with minimal changes.  This can then be run as any user `docker run -d --user 1000 new-image`.
```
FROM nginx
RUN chmod 777 /var/cache/nginx \
	&& sed -i -e '/listen/!b' -e '/80;/!b' -e 's/80;/8080;/' /etc/nginx/conf.d/default.conf \
	&& sed -i 's!/var/run/nginx.pid!/var/cache/nginx/nginx.pid!g' /etc/nginx/nginx.conf
```

On a new enough kernel (4.11+) you can even skip the port 80 swapping and run it with `--sysctl net.ipv4.ip_unprivileged_port_start=0`. With some testing by @tianon, you can run the current image as-is using any user. by working around the two directories that need write access, and the sysctl to allow binding to port 80.

```bash
docker run -it --rm --user 1000:1000 \
--sysctl net.ipv4.ip_unprivileged_port_start=0 \
--mount type=volume,destination=/var/cache/nginx,volume-opt=type=tmpfs,volume-opt=device=tmpfs \
--mount type=volume,destination=/var/run,volume-opt=type=tmpfs,volume-opt=device=tmpfs \
nginx
```

mainline/openshift/Dockerfile
tobru: See https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-origin-specific-guidelines -> Support Arbitrary User IDs

alessfg:
> docker run -it --rm --user 1000:1000 \
--sysctl net.ipv4.ip_unprivileged_port_start=0 \
--mount type=volume,destination=/var/cache/nginx,volume-opt=type=tmpfs,volume-opt=device=tmpfs \
--mount type=volume,destination=/var/run,volume-opt=type=tmpfs,volume-opt=device=tmpfs \
nginx

The current kernel employed by OpenShift is 4.9 LTS. I tested using --sysctl just in case but no luck.

> FROM nginx
RUN chmod 777 /var/cache/nginx \
	&& sed -i -e '/listen/!b' -e '/80;/!b' -e 's/80;/8080;/' /etc/nginx/conf.d/default.conf \
	&& sed -i 's!/var/run/nginx.pid!/var/cache/nginx/nginx.pid!g' /etc/nginx/nginx.conf

On the other hand, this works and lets you run NGINX as non-root. However, to get this image running within OpenShift I still need to push it to Dockerhub.

> We support being able to run an image as non-root and many service images support running as any arbitrary user, like postgres, but we resist changes that are meant only for a specific environment like docker-library/docker#12.

That makes sense. Would shifting the focus from OpenShift to a non-root Docker image be OK then? A quick Google search reveals some interest in running NGINX as non-root in various environments.

yosifkit:
> That makes sense. Would shifting the focus from OpenShift to a non-root Docker image be OK then? A quick Google search reveals some interest in running NGINX as non-root in various environments.

From the official-images team we would have no objections to this.  We just wanted to make sure that an environment specific image doesn't pop up.  It is up to the Nginx maintainer on whether to support running the image as an arbitrary user. 

On a side note you could look at automated builds (https://docs.docker.com/docker-hub/builds/) and repository links (https://docs.docker.com/docker-hub/builds/#repository-links) to have an up-to-date image on the Docker Hub using my example Dockerfile.

alessfg:
Cool. I'll create another PR with a different approach 😄

PR #193, opened 2017-08-31
Add support for s390x via "apt-get source --compile"

tianon:
This takes advantage of the fact that http://nginx.org/packages/ includes the source packages as well, and simply compiles them on non-prebuilt architectures instead, which ensures we have a completely compatible experience with minimal maintenance overhead.

This is based heavily on https://github.com/docker-library/postgres/pull/330, and I've tried to be a little bit over-the-top on comments. :innocent:

I've personally tested all 4 builds on both `amd64` and `s390x`. :+1:

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -65,6 +65,7 @@ for version in "${versions[@]}"; do
 	echo
 	cat <<-EOE
 		Tags: $(join ', ' "${versionAliases[@]}")
+		Architectures: amd64, i386, s390x
 		GitCommit: $commit
 		Directory: $version/$base
 	EOE
@@ -78,6 +79,7 @@ for version in "${versions[@]}"; do
 		echo
 		cat <<-EOE
 			Tags: $(join ', ' "${variantAliases[@]}")
+			Architectures: amd64, i386, s390x
 			GitCommit: $commit
 			Directory: $version/$variant
 		EOE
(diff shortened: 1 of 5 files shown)


generate-stackbrew-library.sh
```diff
@@ -65,6 +65,7 @@ for version in "${versions[@]}"; do
 	echo
 	cat <<-EOE
 		Tags: $(join ', ' "${versionAliases[@]}")
+		Architectures: amd64, i386, s390x
```
tianon: I've currently hard-coded this minimal `Architectures` list, but as-is, these updates should technically support several other architectures including `arm64v8`, `arm32v7`, and `ppc64le`.  For other official images, we usually use `bashbrew` to scrape the list of `Architectures` directly from the parent image to embed in the children for cases like this, but I figured that's something we ought to discuss before I implemented it. :innocent: :heart:

I'd also be happy to simply adjust this list to include more arches if you'd rather, but then they'll need to be manually updated for any base image changes or new arches (which may be desirable).

I'm also happy to test on any other arches. :+1:

thresheek: I'm fine with providing images for all the other arches supported - we've seen requests for armv7 and ppc64le at least.

LGTM otherwise.

Thanks!

tianon: Alright, neato!  I'll have this updated shortly!

tianon:
I've successfully built `stable/stretch` on `ppc64le`, `arm64v8`, and `arm32v7`.  Happy to do more build tests, but also happy to PR again afterwards if the official images builds happen to fail for some odd reason.

thresheek:
Many thanks!

PR #196, opened 2017-09-14
Add support for non-root NGINX Docker images

alessfg:
Initial implementation of Dockerfiles to create a set of NGINX Docker images that can be run as any user besides root (based on the discussion in nginxinc/docker-nginx#188)

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -85,6 +85,21 @@ for version in "${versions[@]}"; do
 		EOE
 	done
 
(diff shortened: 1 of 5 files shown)


kostyrev:
Will this ever be merged?

alessfg:
Hi @kostyrev! You can now find an NGINX Docker non-root repository here (https://github.com/nginxinc/docker-nginx-openshift) and Docker images here (https://hub.docker.com/r/nginxinc/openshift-nginx/).

PR #200, opened 2017-09-22
Adjust scanelf to ignore local libs

yosifkit:
Copying the improvement from https://github.com/docker-library/ruby/pull/161.  Should be no discernible change in packages installed, but does make the sub-shell a little easier to understand.


thresheek:
Merged, thanks!

PR #206, opened 2017-10-15
Use TLS to fetch nginx sources, signatures and public keys

J0WI:


thresheek:
We're checking sources against a known signature, so TLS is absolutely unneeded here.

Also, at least keyserver.ubuntu.com does not speak hkps on port 80 - did you actually check what you've committed?

J0WI:
There was already discussion [here](https://github.com/docker-library/mysql/pull/212#issuecomment-246508858) about the benefits of TLS. In the case of http I see *no* drawback.
Meanwhile, using https has become a common practice for official images:

https://github.com/docker-library/golang/pull/12
https://github.com/docker-library/drupal/pull/55
https://github.com/docker-library/ruby/pull/87
https://github.com/docker-library/redmine/pull/34
https://github.com/docker-library/php/pull/293
https://github.com/docker-library/busybox/commit/8b0d96682de7f70e9a67bc9ab1a488da289ede38
https://github.com/docker-library/memcached/commit/0991958bb39709ddec95c96e23cca863196e605e


I tested the build locally and it works for me. If you prefer, I can split the hkps change into another PR. I never seen Dockerfiles using multiple key servers, but I also didn't found a guideline about this topic.

thresheek:
It works because it uses the first gpg keyserver to fetch the keys.  It does indeed seems that https is becoming prevalent in official images, so fix that PR and re-send.

J0WI:
@thresheek I've created #208.

PR #211, opened 2017-11-10
Upd: Boost worker_connections from 1024 to 10240.

cadesalaberry:
> It's very common to have thousands of clients, especially for a public website. I stopped counting the amount of websites I've seen went down because of the low defaults.

[source](https://serverfault.com/questions/787919/optimal-value-for-nginx-worker-connections)


thresheek:
Defaults are just the defaults.  They are never meant to cover all use cases.

If you need to change them for your particular usecase, feel free to do that in your own nginx.conf.

PR #216, opened 2017-12-20
Add brotli Alpine builds

### LINKED ISSUE #210: Brotli support
Hey there,

As brotli has already become supported by almost all the major browsers, it would be awesome if you could include it in the images.

What do you think?

abevoelker:
Only added to the Alpine builds. Closes #210


alaz:
Doubtful quality: https://github.com/google/ngx_brotli/issues/62

thresheek:
Nack.  We won't ship third-party modules in the official images since we cant ensure the quality of those.

abevoelker:
Fair enough, I expected that tbh :smile_cat:  Thanks for your consideration.

PR #217, opened 2017-12-22
WIP: add envsubst wrapper

glensc:
the wrapper replaces only known env variables, leaves unknown ones
intact

this addresses problem where unknown variables are replaced empty:
https://github.com/docker-library/docs/issues/496

my change was inspired from this comment:
https://github.com/docker-library/docs/issues/496#issuecomment-313065989

but uses more standard solution instead bash specific `compgen`

### DIFF
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -138,6 +138,7 @@ RUN GPG_KEYS=B0F4253373F8F6F510D42178520A9993A1C052F8 \
 
 COPY nginx.conf /etc/nginx/nginx.conf
 COPY nginx.vh.default.conf /etc/nginx/conf.d/default.conf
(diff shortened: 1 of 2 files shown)


glensc:
this is Work In Progress, not sure how it should be really done properly, wrapper script copied beside each `Dockerfile` variant?

glensc:
@tianon care to take look and give guidelines?

thresheek:
I feel like modyfing the configuration on the fly is not something I'm comfortable with in an official image.  Given the amount of workarounds mentioned here and there to achieve the same idea in the docker-compose or similar solutions, I'm inclined to keep things as is, and maybe document it better.

glensc:
will be merged with https://github.com/nginxinc/docker-nginx/pull/232

thanks @skomma

PR #224, opened 2018-01-26
Added brotli module install for alpine editions

dzek69:
This add Google's Brotli compression module for nginx. Brotli is faster and more effective than gzip on default compression level and for static files it results with around 20% smaller sizes than gzip.

It's supported by all major browsers (except IE, which is not updated anymore, but Chrome, Firefox, Edge, Safari, they all support Brotli) and of course nginx will use it only for browsers that supports it.

It's around for few years and Facebook & Google (as example of the biggest ones) are serving their text-based assets using Brotli. Having this in most popular docker image of nginx would definitely help spreading it's usage.

Debian images are to-be-done. I did it mostly for myself and wanted to see if this even got accepted. I'll add it for Debian-based images as well if you're going to accept this.

### DIFF
--- a/mainline/alpine-perl/Dockerfile
+++ b/mainline/alpine-perl/Dockerfile
@@ -50,7 +50,15 @@ RUN GPG_KEYS=B0F4253373F8F6F510D42178520A9993A1C052F8 \
 		--with-compat \
 		--with-file-aio \
 		--with-http_v2_module \
+		--add-module=/usr/local/src/ngx_brotli \
 	" \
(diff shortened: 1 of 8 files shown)


thresheek:
No, it won't be accepted as we would not ship third party code in nginx official images.  Sorry about that.

Also if you're going to use it in your own images, I would say that checking out non-specific git revision will lead to troubles.

dzek69:
Ok, no problem, seems resonable.

Thanks for pointing out checking out non-specific branch, I forgot to update it.

Issue #230, opened 2018-02-21
no matching manifest for linux/amd64 in the manifest list entries


kelaductor:
in Dockerfile 
```
FROM nginx:alpine
```

```
$ docker-compose build
Building nginx
Step 1/5 : FROM nginx:alpine
alpine: Pulling from library/nginx
ERROR: Service 'nginx' failed to build: no matching manifest for linux/amd64 in the manifest list entries
```

uLan08:
same here
`docker pull nginx:1.13-alpine`

```
1.13-alpine: Pulling from library/nginx
no matching manifest for linux/amd64 in the manifest list entries
```

uLan08:
I saw in other related issues that this should fix itself after a few mins but it has been an hour since then. 
Is it possible to trigger a manual build right now? @tianon 

Tagging for future reference 
https://github.com/docker-library/official-images/issues/3835

kb10uy:
Same here too.
```yaml
web:
  image: nginx:mainline-alpine
```
I got
```
Pulling web (nginx:mainline-alpine)...
mainline-alpine: Pulling from library/nginx
ERROR: no matching manifest for linux/amd64 in the manifest list entries
```
Like @uLan08 said, this image seems to be down for long time than other cases.
I hope this recover asap. 🙏

mattjegan:
While the image is down you can possibly get around this issue temporarily using an older image, e.g. `nginx:1.12-alpine`

TopFuel:
Until this is fixed you can use `nginx:stable-alpine`, it has an older version of nginx, but it's working fine for me at the moment.

uLan08:
A better approach would probably be too specifically pull an old sha256 digest of nginx:1.13-alpine or whatever tag you're using. This way you won't be pulling an old version of nginx rather just an old build. It is also the recommended workaround in the issue I have linked. 
What I did was I viewed the history of the [nginx alpine readme](https://github.com/docker-library/repo-info/blob/master/repos/nginx/remote/1.13-alpine.md) copied and pulled the old digest and tagged it as `nginx:1.13-alpine`

rawpixel-vincent:
you can also use the previous build directly 
`FROM nginx:1.13.8-alpine`

kennyrulez:
@rawpixel-vincent your snippet did the trick for me. Thanks

ebernard:
See also https://github.com/docker-library/official-images/issues/3835

thresheek:
[...]

PR #231, opened 2018-02-21
Add "-f" to GNUPGHOME deletions (helps avoid a GnuPG 2.x race condition)

tianon:
This is the race condition that has been plaguing the official build for the past many hours, and we've found that adding "-f" helps mitigate it (doesn't completely eliminate it, but it's a dramatic improvement in the successful build ratio).


tianon:
(See https://github.com/nginxinc/docker-nginx/issues/230)

PR #232, opened 2018-03-03
Run envsubst automatically during starting container

skomma:
This PR adds the feature to automatically run envsubst for files with the suffix `.conf.template` on `/etc/nginx`, and output to `.conf` file.
If there is no file with the suffix `.conf.template`, the added script does nothing.

This PR was made with reference to PR #217. 
Thanks @glensc!

### DIFF
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -110,6 +110,10 @@ RUN set -x \
     && ln -sf /dev/stdout /var/log/nginx/access.log \
     && ln -sf /dev/stderr /var/log/nginx/error.log
 
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+
 EXPOSE 80
 
 STOPSIGNAL SIGTERM
--- a/mainline/alpine/entrypoint.sh
+++ b/mainline/alpine/entrypoint.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+set -ex
+
+main() {
+  local template_dir="${NGINX_ENVSUBST_TEMPLATE_DIR:-/etc/nginx/templates}"
+  local suffix="${NGINX_ENVSUBST_TEMPLATE_SUFFIX:-.template}"
+  local output_dir="${NGINX_ENVSUBST_OUTPUT_DIR:-/etc/nginx/conf.d}"
+
+  local template defined_envs relative_path output_path subdir
+  defined_envs=$(printf '${%s} ' $(env | cut -d= -f1))
+  [ -d "$template_dir" ] || return 0
+  [ -w "$output_dir" ] || return 0
+  for template in $(find "$template_dir" -follow -name "*$suffix"); do
+    relative_path="${template#$template_dir/}"
+    output_path="$output_dir/${relative_path%$suffix}"
+    subdir=$(dirname "$relative_path")
+    # create a subdirectory where the template file exists
+    mkdir -p "$output_dir/$subdir"
+    envsubst "$defined_envs" < "$template" > "$output_path"
+  done
+}
+
+main
+
+exec "$@"
--- a/mainline/alpine/entrypoint.sh
+++ b/mainline/alpine/entrypoint.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-main() {
+auto_envsubst() {
   local template_dir="${NGINX_ENVSUBST_TEMPLATE_DIR:-/etc/nginx/templates}"
   local suffix="${NGINX_ENVSUBST_TEMPLATE_SUFFIX:-.template}"
   local output_dir="${NGINX_ENVSUBST_OUTPUT_DIR:-/etc/nginx/conf.d}"
@@ -10,7 +10,10 @@ main() {
   local template defined_envs relative_path output_path subdir
   defined_envs=$(printf '${%s} ' $(env | cut -d= -f1))
   [ -d "$template_dir" ] || return 0
-  [ -w "$output_dir" ] || return 0
+  if [ ! -w "$output_dir" ]; then
+    echo "ERROR: $template_dir exists, but $output_dir is not writable. exiting..." 1>&2
+    return 1
+  fi
   for template in $(find "$template_dir" -follow -name "*$suffix"); do
     relative_path="${template#$template_dir/}"
     output_path="$output_dir/${relative_path%$suffix}"
@@ -21,6 +24,6 @@ main() {
   done
 }
 
-main
+auto_envsubst
 
 exec "$@"


glensc:
similarly to #217 this addresses only alpine image, and #217 was not finished because received no feedback how.

thresheek:
I feel like modyfing the configuration on the fly is not something I'm comfortable with in an official image.  Given the amount of workarounds mentioned here and there to achieve the same idea in the docker-compose or similar solutions, I'm inclined to keep things as is, and maybe document it better.

skomma:
@thresheek Thank you for comment.

> Given the amount of workarounds mentioned here and there to achieve the same idea in the docker-compose or similar solutions,

I think the official image should have this feature since there are the amount of workarounds all around the Internet.
And this PR's approach affects **ONLY** files with `.conf.template` suffix.

If you agree this, I'll write how to use in README, and modify all kinds of Dockerfiles!

mainline/alpine/nginx-envsubst.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
```
glensc: add `set -e`, so the container startup be aborted in case processing error

glensc:
as author of the original PR https://github.com/nginxinc/docker-nginx/pull/217, i agree that official image should provide working solution, as there are multiple solutions in the wild, some of them have flaws and side effects.

having it here provides single truth and common sense.

thresheek:
Fair enough - please update the PR with the README/Dockerfiles.

FYI, I'll be on vacation until mid-may so I wouldnt be able to review it until then. TIA.

skomma:
@thresheek Thank you! I'll update this PR.

glensc:
@skomma https://github.com/nginxinc/docker-nginx/pull/217 added standalone wrapper that could be used to replace any file, would it make sense to keep that support around?

I think there may be setups around that need the subst wrapper but don't match with your defined patterns.

or it's better to have only one pattern, not to support variety of usecases?

skomma:
@glensc Thank you for comment.

>
I think it's better that the target files of this PR are only nginx
configuration files.
This behavior is easy to document, and is easy to understand.

Do you have any other use cases?

glensc:
@skomma I mean if for example my legacy (prior this PR) template are not in `/etc/nginx` dir, or I use different extension than `.conf.template`:

- /opt/app/config/app.conf.tpl -> /etc/nginx/conf.enabled/app.conf

skomma:
@glensc I see.

But once this image decide a rule of target templates, will you place the files in `/etc/nginx/conf.d` with `.conf.template` suffix?

glensc:
@skomma I don't have problem following the pattern, just thinking out loud for scenarios where can't for some reason follow the rules (have own setup)

skomma:
@glensc OK, I'll implement to change the settings of target files via env vars, and make "find" follow symlinks.
I believe that this will work for most use cases.
(directories can be changed using "-v" option in docker)

If the users want to set more complex settings, I think that it is good in the design to have their own settings.(ex. using their own scripts)

skomma:
@glensc I think I've fixed the feedback point from you. (https://github.com/nginxinc/docker-nginx/commit/a42f68d51b34e9f8affdacdd8bb860c57a9e419e#commitcomment-33368429)
Is it as you expected?

glensc:
i wrote two comments regarding that mkdir:
- https://github.com/nginxinc/docker-nginx/commit/a42f68d51b34e9f8affdacdd8bb860c57a9e419e#commitcomment-33368429
- the one in thread itself. can't find the link

but you totally misunderstood me. i said do not add your line as **last** of the `RUN` statement, but **anywhere** else. but you instead added another layer to docker image by adding new `RUN`.

perhaps this explains that better:

no:
```diff
RUN aa \
  && bb \
  && cc \
-  && dd
+  && dd \
+  && new code
```

yes:
```diff
RUN aa \
  && bb \
+  && new code \
  && cc \
  && dd
```

skomma:
> * the one in thread itself. can't find the link

https://github.com/nginxinc/docker-nginx/commit/0d7275decf92f7600ca0b477730965e62736826d#r33357557

> but you totally misunderstood me. i said do not add your line as **last** of the `RUN` statement, but **anywhere** else. but you instead added another layer to docker image by adding new `RUN`.

Sorry...Certainly, I couldn't understand what you said.

But I can not agree the location of `mkdir` you suggested.
I can not understand to write in the middle of the existing RUN statement, that is the core installation of nginx.
`nginx-envsubst` is a **completely** additional component, so it is not bad to separate `mkdir` RUN statement from the existing one, I think.

glensc:
I'm not an official reviewer, just gave my thoughts. Minimizing layers is one of the best practices when building docker images. Avoiding formatting diffs is best practices for general development.

- https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

I personally would use multi-layer builds to minimize layers, don't know why that practice is not used in official images.

example, adds only 1 layer: https://gitlab.com/gitlab-org/cluster-integration/helm-install-image/merge_requests/5/diffs

skomma:
I knew it was generally better that a docker image has a smaller number of layers in a docker image.
But in this case, the RUN statement only creates a directory, not including deletion, and I think it would be better to separate since it doesn't relate to the nginx core components.

This PR is now ready to be reviewed, I think.

@thresheek Could you review this and the document PR docker-library/docs#1471?

thresheek:
I think it makes more sense to run the envsubst as a part of entrypoint, what do you think?  I dislike calling shell in the CMD. What I also dislike is an avoidable layer, no need to create /etc/nginx/template in the Dockefile - it should be documented in the docs instead.

There is also a problem if the container is ran as an unprivileged user, we will have no way to output to /etc/nginx/conf.d.  The attached patch doesnt solve the issue, but checks if we can write to that path. I guess documenting that is a good idea too.

Here a patch (against current master, b749353968a57ebd9da17e12d23f1a5fb62f9de9)
```
diff --git a/mainline/alpine/Dockerfile b/mainline/alpine/Dockerfile
index 17822da..c8b0dbd 100644
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -110,6 +110,10 @@ RUN set -x \
     && ln -sf /dev/stdout /var/log/nginx/access.log \
     && ln -sf /dev/stderr /var/log/nginx/error.log

+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+
 EXPOSE 80

 STOPSIGNAL SIGTERM
diff --git a/mainline/alpine/entrypoint.sh b/mainline/alpine/entrypoint.sh
new file mode 100755
index 0000000..2620d8c
--- /dev/null
+++ b/mainline/alpine/entrypoint.sh
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+set -ex
+
+main() {
+  local template_dir="${NGINX_ENVSUBST_TEMPLATE_DIR:-/etc/nginx/templates}"
+  local suffix="${NGINX_ENVSUBST_TEMPLATE_SUFFIX:-.template}"
+  local output_dir="${NGINX_ENVSUBST_OUTPUT_DIR:-/etc/nginx/conf.d}"
+
+  local template defined_envs relative_path output_path subdir
+  defined_envs=$(printf '${%s} ' $(env | cut -d= -f1))
+  [ -d "$template_dir" ] || return 0
+  [ -w "$output_dir" ] || return 0
+  for template in $(find "$template_dir" -follow -name "*$suffix"); do
+    relative_path="${template#$template_dir/}"
+    output_path="$output_dir/${relative_path%$suffix}"
+    subdir=$(dirname "$relative_path")
+    # create a subdirectory where the template file exists
+    mkdir -p "$output_dir/$subdir"
+    envsubst "$defined_envs" < "$template" > "$output_path"
+  done
+}
+
+main
+
+exec "$@"
```

Let me know what you think.

skomma:
Thank you for review and diff.

> I think it makes more sense to run the envsubst as a part of entrypoint, what do you think? I dislike calling shell in the CMD.

Oh, the idea of ​​leaving the command as-is and preparing an entrypoint script did not come to my mind!
That's the good idea. I agree this at all!

> What I also dislike is an avoidable layer, no need to create /etc/nginx/template in the Dockefile - it should be documented in the docs instead.

Ok, I'll document that the users must prepare the template directory as `/etc/nginx/template`, if they want to use this method.

> There is also a problem if the container is ran as an unprivileged user, we will have no way to output to /etc/nginx/conf.d. The attached patch doesnt solve the issue, but checks if we can write to that path. I guess documenting that is a good idea too.

Ok, I'll also document such permission issues.

skomma:
@thresheek I fixed about [this](#issuecomment-499311489).

Would you review this and the related document PR docker-library/docs#1471?

glensc:
@thresheek cheers for finally merging this in cd98e75!

PR #246, opened 2018-05-10
Added '--with-ipv6' flags to alpine and alpine-perl images

bjrnio:


thresheek:
NGINX configure does not have that configure option anymore, see http://hg.nginx.org/nginx/rev/a6d116645c51.

What are you trying to solve?

bjrnio:
Well I guess it's just a misconfiguration and some coincidences. An image I used with --with-ipv6 had worked for me in the past over IPv6 and now I'm using the vanilla one and it doesn't, but if you say IPv6 is enabled by default now I guess it's just a misconfiguration of mine. I'll close.

PR #255, opened 2018-07-10
Bump alpine to 3.8

michael-k:


vurt007:
any chance this PR can be accepted? considering this-> https://justi.cz/security/2018/09/13/alpine-apk-rce.html

thresheek:
apk security update went into 3.7 branch as well.

J0WI:
Not always. Or at least not always immediately. E.g. CVE-2018-14618 (curl) or CVE-2018-12020 (gnupg) are both fixed in 3.8, but have not yet been backported to 3.7.

artheus:
👍 I think that it would be a good idea to bump alpine version to 3.8 either way. It is the latest version, and is a supported release. I cloned this repository and built nginx with alpine:3.8 as base for nginx and it seems to work perfectly, without other changes to the Dockerfile (I only tested the non-perl Dockerfile, though)

My current solution will now be to fork this repository and use 3.8 as base and store the nginx base image in my private docker repo.

I can fork this repo in Github as-well and send a pull-request for the alpine images if that would help to get this change through faster.

J0WI:
Since most other alpine based images have already been upgraded to `3.8` this will also enhance caching, disk usage and download times. This is because container layers can be shared between images.

thresheek:
Mainline images starting from 1.15.4-alpine will be based on 3.8.

michael-k:
Superseded by commit 3c446adddf7b04e877d5c5191b151c794c74bb58

PR #274, opened 2018-09-19
Upgrade to GnuPG 2

J0WI:
This resolves the issue described in https://github.com/nginxinc/docker-nginx/commit/f3fc4d5753f0ebb9107738183b9c5cea1bf3f618.


thresheek:
I dont really see the benefits here?

thresheek:
Exactly, so it's supported security-wise and does everything we need already. So no benefits to upgrade.

Issue #218, opened 2018-01-03
Inconsistent user ids


colinmollenhour:
The nginx config specifies the user "nginx", but this user does not exist on a clean base installation for the host OS so Docker maps it to some other user id. Somehow between 1.10 and 1.12 this seems to have changed from 104 to 101 so that caused file permission errors because 101 was "systemd-network" on the host and 104 was "syslog" (Ubuntu host).

Is there some way to guaranatee the uid for "nginx" doesn't change without having to add users directly on the host, since having to do this defeats the purpose of Docker.

Perhaps a simple solution would be for the docker entrypoint to chown the /var/cache/nginx and /var/log/nginx directories to "nginx"?

tianon:
> this user does not exist on a clean base installation for the host OS so Docker maps it to some other user id

This isn't exactly accurate.  Docker doesn't do any mapping at all.  The Linux kernel doesn't deal in user _names_ at all, and only cares about IDs -- names are a userspace abstraction to help us humans remember and conceptually deal with the IDs.

```console
$ docker run -it --rm nginx id nginx
uid=101(nginx) gid=101(nginx) groups=101(nginx)
```

So, even if your host has an `nginx` user that's got an ID of (for example) 1000, this container will still by default run as UID 101 (which may or may not match anything in your host's `/etc/passwd` file).

If you have a new enough kernel, something like https://github.com/moby/moby/issues/8460#issuecomment-312459310 can be used to run the `nginx` image as an arbitrary user, but otherwise you can adjust the configuration file to use a port other than 80.  Here's a working example to run `nginx:latest` as UID 1000:

```console
$ docker run -it --rm --user 1000:1000 --tmpfs /var/cache/nginx:uid=1000 --tmpfs /run:uid=1000 --sysctl net.ipv4.ip_unprivileged_port_start=0 nginx
2018/01/03 19:06:05 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
```

colinmollenhour:
[...]

PR #280, opened 2018-10-05
Use stable gid/uid

### LINKED ISSUE #113: Permissions issue (using compose)
A simple compose configuration leaves my static files owned by the UID of my host user, and inaccessible to the nginx process.

```yaml
nginx:
    image: nginx:stable-alpine
    volumes:
        - ./app:/usr/share/nginx/html
    ports:
        - '80:80'
        - '443:443'
```

The directory inside the container is, in this case, owned by `1000`, which is the user's ID from my host machine. `./app/` simply contains a single `index.html` file.

I don't see any options in [the documentation](https://github.com/docker-library/docs/tree/master/nginx) that cover permissions. Is there an environment variable I can submit to the container during build to change the UID of the www-data user?

artem-zinnatullin:
Fixes #113, specifically https://github.com/nginxinc/docker-nginx/issues/113#issuecomment-237482535.

### Motivation

Nginx proxy_cache and other features need read/write access to the disk. Disk normally is mounted upfront and gets fs mounted and owned by `root` who runs docker/etc daemon.

For Nginx to be able to read/write mounted disk, files **must** have proper access permissions.

One of the best ways to do it is to assign ownership to the proper user/group upfront. 

Unfortunately, Nginx atm doesn't use fixed gid/uid (they're fixed within Docker image version, but can be changed on next image rebuild due to changes in Alpine) and this complicates setup a lot, especially in environments like Kubernetes where you can't control external gid/uid or pass it to the container at startup.

This PR adds fixed high-order gid/uid values to address that.

Similar change in couchdb for reference: https://github.com/apache/couchdb-docker/pull/64/files

With this change it'll be possible to mount disk like this (k8s):

```yaml
initContainers:
- name: "grant-permissions"
  image: "busybox:1.26.2"
  env:
  - name: NGINX_GID
    value: 8123
  - name: NGINX_UID
    value: 8123
  command:
  - 'sh'
  - '-c'
  - 'chown -R ${NGINX_UID}:${NGINX_GID} /opt/nginx-cache'
  volumeMounts:
  - mountPath: "/opt/nginx-cache"
    name: nginx-cache
containers:
- name: nginx
  image: nginx:1.15.4-alpine
  volumeMounts:
  - name: nginx-cache
    mountPath: /opt/nginx-cache
```

---

If this change gets approved in general, I'll update the remaining `Dockerfile`s in the repo in this PR

### DIFF
--- a/mainline/stretch/Dockerfile
+++ b/mainline/stretch/Dockerfile
@@ -5,6 +5,10 @@ LABEL maintainer="NGINX Docker Maintainers <docker-maint@nginx.com>"
 ENV NGINX_VERSION 1.15.5-1~stretch
 ENV NJS_VERSION   1.15.5.0.2.4-1~stretch
 
+# Stable gid/uid allows you pre-configure disk volumes with proper ownership (`chown`) for use by Nginx.
+ENV NGINX_GID 8123
+ENV NGINX_UID 8123
+
 RUN set -x \
 	&& apt-get update \
 	&& apt-get install --no-install-recommends --no-install-suggests -y gnupg1 apt-transport-https ca-certificates \
@@ -87,6 +91,9 @@ RUN set -x \
 		apt-get purge -y --auto-remove \
 		&& rm -rf "$tempDir" /etc/apt/sources.list.d/temp.list; \
 	fi
+
+RUN groupmod -g $NGINX_GID nginx \
+    && usermod -u $NGINX_UID nginx
 
 # forward request and error logs to docker log collector
 RUN ln -sf /dev/stdout /var/log/nginx/access.log \
--- a/mainline/stretch/Dockerfile
+++ b/mainline/stretch/Dockerfile
@@ -5,10 +5,6 @@ LABEL maintainer="NGINX Docker Maintainers <docker-maint@nginx.com>"
 ENV NGINX_VERSION 1.15.5-1~stretch
 ENV NJS_VERSION   1.15.5.0.2.4-1~stretch
 
-# Stable gid/uid allows you pre-configure disk volumes with proper ownership (`chown`) for use by Nginx.
-ENV NGINX_GID 8123
-ENV NGINX_UID 8123
-
 RUN set -x \
 	&& apt-get update \
 	&& apt-get install --no-install-recommends --no-install-suggests -y gnupg1 apt-transport-https ca-certificates \
@@ -92,9 +88,6 @@ RUN set -x \
 		&& rm -rf "$tempDir" /etc/apt/sources.list.d/temp.list; \
 	fi
 
-RUN groupmod -g $NGINX_GID nginx \
-    && usermod -u $NGINX_UID nginx
-
 # forward request and error logs to docker log collector
 RUN ln -sf /dev/stdout /var/log/nginx/access.log \
 	&& ln -sf /dev/stderr /var/log/nginx/error.log
@@ -102,5 +95,8 @@ RUN ln -sf /dev/stdout /var/log/nginx/access.log \
 EXPOSE 80
 
 STOPSIGNAL SIGTERM
+
+COPY entrypoint.sh /opt/entrypoint.sh
+ENTRYPOINT ["/opt/entrypoint.sh"]
 
 CMD ["nginx", "-g", "daemon off;"]
--- a/mainline/alpine-perl/Dockerfile
+++ b/mainline/alpine-perl/Dockerfile
@@ -4,6 +4,11 @@ LABEL maintainer="NGINX Docker Maintainers <docker-maint@nginx.com>"
 
 ENV NGINX_VERSION 1.15.5
 
+# Stable gid/uid allows you pre-configure disk volumes with proper ownership (`chown`) for use by Nginx.
+ENV NGINX_GID 8123
+ENV NGINX_UID 8123
+
+
 RUN GPG_KEYS=B0F4253373F8F6F510D42178520A9993A1C052F8 \
 	&& CONFIG="\
 		--prefix=/etc/nginx \
@@ -51,8 +56,8 @@ RUN GPG_KEYS=B0F4253373F8F6F510D42178520A9993A1C052F8 \
 		--with-file-aio \
 		--with-http_v2_module \
 	" \
-	&& addgroup -S nginx \
-	&& adduser -D -S -h /var/cache/nginx -s /sbin/nologin -G nginx nginx \
+	&& addgroup -S --gid $NGINX_GID nginx \
+	&& adduser -D -S -h /var/cache/nginx -s /sbin/nologin -G nginx --uid $NGINX_UID nginx \
 	&& apk add --no-cache --virtual .build-deps \
 		gcc \
 		libc-dev \
--- a/mainline/alpine-perl/Dockerfile
+++ b/mainline/alpine-perl/Dockerfile
@@ -4,11 +4,6 @@ LABEL maintainer="NGINX Docker Maintainers <docker-maint@nginx.com>"
 
 ENV NGINX_VERSION 1.15.5
 
-# Stable gid/uid allows you pre-configure disk volumes with proper ownership (`chown`) for use by Nginx.
-ENV NGINX_GID 8123
-ENV NGINX_UID 8123
-
-
 RUN GPG_KEYS=B0F4253373F8F6F510D42178520A9993A1C052F8 \
 	&& CONFIG="\
 		--prefix=/etc/nginx \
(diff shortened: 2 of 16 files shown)


thresheek:
This cant be merged since it only applies to alpine - to move forward we'd need to have the same on all supported variants.

thresheek:
It makes sense, but given we use prebuilt debian packages for non-alpine builds, it'd be hard to merge it.

thresheek:
I mean I'm not sure we can somehow delay the pre-inst stage on debian packages install so we fix it in the Dockerfile.

KostyaEsmukov:
I believe this change being published under the currently existing docker tags might break a number of installations, which already rely on the current uid:gid pair. Could you consider publishing this under new tags or make this behavior optional somehow (e.g. with a script in the entrypoint)?

artem-zinnatullin:
@thresheek I've updated PR with gid/uid change in Debian based images (and rest of Dockerfiles)

@KostyaEsmukov that is a good point. 
I've thought about it for couple days now and came to conclusion that while this is a change, it's hard to consider it breaking because nothing guaranteed stable gid/uid in the first place. If someone relied on prev gid/uid values, it should be expected for them to check for change every time when they update the image.

We **must** however indicate it in release notes

KostyaEsmukov:
> it's hard to consider it breaking because nothing guaranteed stable gid/uid in the first place

But they *were* stable (at least on debian images), so actually I *would* consider this a breaking change. Honestly, I don't think I ever witnessed the change in `uid:gid` myself. I suppose that might have happened during upgrade to alpine 3.8, but this is rather a one-time regression than a regular occurrence.

> If someone relied on prev gid/uid values, it should be expected for them to check for change every time when they update the image.

While this is a pretty logical conclusion, I don't think those who *do* rely on the specific `uid:gid` actually did consider that they might change. As I said, I haven't ever seen the change on the debian images.

> We must however indicate it in release notes

I'm afraid not a lot of people do read them before pulling the latest image associated with a tag, especially given the ubiquity of automated deployment tools.

I'm not sure if this is a serious issue at all, but I would still suggest to not introduce it by default in the current tags. I totally agree that having fixed `uid:gid` is the way to go, but I feel that in this form this is more harm than good.

mainline/stretch/Dockerfile
```diff
@@ -88,6 +92,9 @@ RUN set -x \
 		&& rm -rf "$tempDir" /etc/apt/sources.list.d/temp.list; \
 	fi
 
+RUN groupmod -g $NGINX_GID nginx \
+    && usermod -u $NGINX_UID nginx
```
KostyaEsmukov: This won't change the `uid:gid` pair of the files owned by this user, so if there are such files, they won't be owned by the `nginx` user and group anymore.

I don't think that there're any files owned by `nginx` at this point in the image, but I think there should be a comment at least, as this might be not so obvious.

artem-zinnatullin:
Damn it, @KostyaEsmukov, нормально ж сидели :D

okay, I think your arguments are compelling enough to me, especially given the level of stability Nginx aims to provide to its users.

I've updated the PR now with entrypoint that keeps the default behavior as is (with helpful message about what's going on), but now allows to customize gid/uid through environment variable passed at container startup (ie `docker run --env NGINX_GID=1234 --env NGINX_UID=1234`)

Unfortunately, I had to duplicate the `entrypoint.sh` for every `Dockerfile`, because Docker doesn't let reference files out of build dir context. I would have symlinked it otherwise, but it's unclear to me how these images are built by Nginx Inc so I couldn't make such assumptions. Suggestions welcomed!

mainline/stretch/Dockerfile
artem-zinnatullin: Added a note to entrypoint

tianon:
Can't this same thing be accomplished already with just "--user" without
adding any additional environment variables?

(I've used it like that successfully in several places.)

artem-zinnatullin:
@tianon Nginx normally runs worker process(es), along with master process. `--user` will only change master process and will actually affect the functionality of Nginx

```
2018/10/08 16:44:17 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
2018/10/08 16:44:17 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
```

Maybe there is a way to make it work with Nginx, but I'm not aware of it

tianon:
`--user` will change the master, so workers will also change as a result -- they can't be started with more privileges than the master, especially if you add `--security-opt no-new-privileges` (which will enforce that at the kernel layer, which is a much, much firmer guarantee than some environment variables and configuration files could provide inside the image)

Additionally, making `--user` work would help with OpenShift's tendency to run images as arbitrary users.

Combining the following `Dockerfile` with `--sysctl net.ipv4.ip_unprivileged_port_start=0` (in order to allow non-root to bind to port 80; see https://github.com/moby/moby/issues/8460#issuecomment-312459310), I was able to run the resulting image with `--user 1001:1001 --security-opt no-new-privileges` with no issues (even removed `user` from the configuration to silence the warning, but that's not even necessary):

```Dockerfile
FROM nginx

RUN mkdir -p /var/cache/nginx /var/run && chmod 777 /var/cache/nginx /var/run

# remove the (harmless) warning about "user" not making sense / doing anything when running as non-root
RUN sed -ri '/^user[[:space:]]/d' /etc/nginx/nginx.conf
```

Doing `docker top` on the resulting container, I get the following:

```console
$ docker top upbeat_fermi
UID                 PID                 PPID                C                   STIME               TTY                 TIME                CMD
1001                30352               30296               2                   13:11               pts/0               00:00:00            nginx: master process nginx -g daemon off;
1001                30407               30352               0                   13:11               pts/0               00:00:00            nginx: worker process
```

artem-zinnatullin:
Just got hit by this issue switching from `1.15.4-alpine` to `1.15.8` (stretch) and trying to reuse cache saved on disk :(

alpine image had nginx uid `100` and gid `101`, stretch has both as `101` (all of them are pretty much "random" since they're not specified in `Dockerfile`)

@tianon thanks for exploring `--user`, let's move on with your suggestion and remove user directive from `/etc/nginx/nginx.conf`? (we'll need to explore backwards compatibility and implications of that for default use case when container is running with `0` uid and gid)

Kubernetes apparently supports `runAsUser` and `runAsGroup` https://kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups, I was wrong saying that it doesn't (couldn't find it for some reason), my apologies

tianon:
The `user` directive doesn't need to be removed -- the only "harm" while using `--user` with it in place is that you get a warning during NGINX startup (which makes sense). :+1:

artem-zinnatullin:
mhm, right, user can also override default `/etc/nginx/nginx.conf` by mounting their version of the file thus override the `user` directive if needed.

@tianon remaining part is `RUN mkdir -p /var/cache/nginx /var/run && chmod 777 /var/cache/nginx /var/run`

1) Not sure if `777` is a good idea, ideally it would be tied to current user
2) Changing chmod on `/var/run` seems scary, would be great to avoid it
3) If we remove `777`, we probably can make `mkdir -p /var/cache/nginx && chown -R $(id -u):$(id -g) /var/cache/nginx` a part of image entrypoint so we always have these directories with correct ownership

What do you think?

colinmollenhour:
Also related to #218

thresheek:
> In my understanding, running `mkdir -p /var/cache/nginx && chown -R $(id -u):$(id -g) /var/cache/nginx` as part of entrypoint should be ok:
> 
>     * seems to be backwards compatible for current users since they'll just get uid and gid of nginx user/group backed into image as they do right now
> 
>     * doesn't seem to affect container startup performance much as this directory is empty by default and is meant for temporary files so should be ~empty on container restart even if it was persisted
> 
>     * should unblock running container with arbitrary uid/gid set by container runtime

No, when running with an arbitrary user, e.g. docker run --user 300, chown will fail since it's not a root anymore and it doesnt have privileges to change file ownership.

colinmollenhour:
> No, when running with an arbitrary user, e.g. docker run --user 300, chown will fail since it's not a root anymore and it doesnt have privileges to change file ownership.

The chown could be conditional based on if it is running as root.

thresheek:
> > No, when running with an arbitrary user, e.g. docker run --user 300, chown will fail since it's not a root anymore and it doesnt have privileges to change file ownership.
> 
> The chown could be conditional based on if it is running as root.

It wouldnt serve any purpose then: root has already all the rights to create subdirectories as needed for the path we're doing chown for.

colinmollenhour:
I was thinking there was an entrypoint script but I see that is not the case.

I believe it is still an issue if you run the container as root and then set the user in the nginx config. The master process will be root but all of the worker processes will run as the non-root user.

thresheek:
What exactly is the issue?  As a part of nginx startup, master process (running as root) will create any directories needed for nginx workers operation with appropriate rights for non-privileged users.

colinmollenhour:
The issue is when upgrading existing containers that use mounted volumes with existing files/directories. When the UID changes the existing files/directories are no longer writable by the server. This is fixed for future users now by making the uid stable, but it will break for existing users where the old UID does not match the new UID.

thresheek:
with chown $(id -u):$(id -g) it will still be root:root.

colinmollenhour:
Good point, I suppose there is no simple way to deduce the user that the workers will run as without parsing the nginx config or starting the server process or introducing a new environment variable..

PR #287, opened 2018-11-10
Fix #286: added nginx-dav-ext-module

### LINKED ISSUE #286: nginx-dav-ext module missing in alpine image
The Debian based image provides the [nginx-dav-ext module](https://github.com/arut/nginx-dav-ext-module). The Alpine version version lacks this module, which in my view is inconsistent. Can this module be included in the Alpine version as well?

richardbrinkman:
The Debian based image is compiled with the [nginx-dav-ext module](https://github.com/arut/nginx-dav-ext-module) enabled. The Alpine version version is compiled without this module, which in my view is inconsistent.
This merge request adds the nginx-dav-ext module to the Alpine version.


thresheek:
We will not ship third-party modules as a part of base image.  Thank you.

PR #292, opened 2018-11-28
nginx with ubuntu base

smcavallo:
Ubuntu now provides minimal images on dockerhub
https://blog.ubuntu.com/2018/07/09/minimal-ubuntu-released
https://hub.docker.com/_/ubuntu/
There is a strong preference to use ubuntu as the base platform for nginx.
Initial support for dockerized nxing running on ubuntu.

Signed-off-by: S.Cavallo <smcavallo@hotmail.com>


thresheek:
We've discussed adding more OS variants in https://github.com/docker-library/official-images/pull/1462.  Not sure it's worth adding Ubuntu at this point.

PR #300, opened 2019-01-06
Replace openssl by libressl

pando85:


yosifkit:
I would strongly recommend against this: https://github.com/alpinelinux/aports/pull/5406#issuecomment-432247954

pando85:
@yosifkit thanks, I didn't know that. Closing.

PR #320, opened 2019-04-04
alpine: use nginx.org-provided packages.

dirtydoggco19:
For the architectures unavailable on nginx.org, build from the official
packaging sources.

Also adds njs.


thresheek:
No, thanks.

PR #339, opened 2019-06-20
Update Alpine to 3.10

J0WI:


thresheek:
This will fail for x86_64 since we don't ship those packages on nginx.org yet.  Once we do, I'll update the dockerfiles.

ulgens:
Any updates? It's been 12 days.

thresheek:
Fixed with 5abbd4a936e9ebd0bd116511c0dcd5ca27b9f4ad, thanks.

ulgens:
Is there a PR to official-images repo for this change?

thresheek:
No, and there won't be one until a new nginx release is up. I will not change the currently available tags to something new.

ulgens:
Python image manages this by two labels for alpine based images. I'm not sure keeping it on a outdated image is the best solution here.

thresheek:
I don't really see what's outdated with alpine 3.9 since it's a security-supported distro which is not yet EOL.

ulgens:
With "if it works, it's okay" mindset, you're right. But that's the official image program and people, including myself, expect to find best and most up to date images. There are alpine based images, and alpine has 3.10 version but Nginx doesn't have 3.10 based image. Why? Is there any technical issue? Nope. Because someone doesn't want to update it / doesn't see the need.

thresheek:
It is exactly "if it works, don't break it" attitude that made nginx what it is.

Moving from one base OS to another one (albeit a minor release) on the same tag is a disaster waiting to happen - if it works for you it's fine, but surely there are people who rely on specific things inside the current tags.  This is the technical issue with just redoing the tags with the new contents, and this is not something we'll do ever.

Speaking of multiple different distro tags, we've discussed this in https://github.com/docker-library/official-images/pull/1462 and decided to go without it.

PR #375, opened 2019-10-21
Add www-data group & user

xcffl:
As www-data is a widely-used user for upstream server (e.g. in PHP), adding this user into the nginx image helps solving problems regarding permissions. 
For instance, in https://github.com/nextcloud/docker/issues/883, we have to modify the image to fit our needs. If the user is pre-configured, we only need to adjust following the non-root user steps.

For the reason why its uid/gid is 82 (alpine only), see here: https://github.com/docker-library/php/blob/9b073b8852132f386d848e93105c62ae2c73c0fd/7.3/alpine3.9/fpm/Dockerfile#L35

### DIFF
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -10,6 +10,9 @@ RUN set -x \
 # create nginx user/group first, to be consistent throughout docker variants
     && addgroup -g 101 -S nginx \
     && adduser -S -D -H -u 101 -h /var/cache/nginx -s /sbin/nologin -G nginx -g nginx nginx \
+    # ensure www-data user exists
+	&& addgroup -g 82 -S www-data; \
+	&& adduser -u 82 -D -S -G www-data www-data \
     && apkArch="$(cat /etc/apk/arch)" \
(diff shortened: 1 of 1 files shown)


thresheek:
Uh, left the comment in a wrong repo, going to do it again here:

Why not just chmod 755 /var/www/ ?

Running nginx workers as the same uid as php-fpm isnt needed at all.

xcffl:
> Uh, left the comment in a wrong repo, going to do it again here:
> 
> Why not just chmod 755 /var/www/ ?
> 
> Running nginx workers as the same uid as php-fpm isnt needed at all.

I guess because nextcloud [recommends](https://docs.nextcloud.com/server/stable/admin_manual/maintenance/manual_upgrade.html?highlight=chmod) so? Probably for security.

thresheek:
Closing this MR since there is no fix required on nginx side for it to work.

xcffl:
It's definitely unwise to have 755 on your php files exposed to the internet. For those who encounter this problem, you may use bitnami's container.

Issue #421, opened 2020-06-05
Verbose entrypoint


alexandresalome:
Since a last month, verbosity was added to the container, making it log on output some information when it starts:

```
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
```

While it's great to have this information in development, I prefer to have information hidden in production, to keep my production logs for only warning/error messages.

Would you agree if I add a `ENTRYPOINT_SHOW_INFO` environment variable or some other mechanism (indicate me) to reduce this verbosity when everything goes as planned?

Default would be to show, as it is now, but allow users to hide it when they want.

thresheek:
That sounds like a good feature to have, indeed.  An environment variable should be fine.

tianon:
There are quite a few `echo` statements to cover -- instead of making them all conditional, what do you think about something like this in `docker-entrypoint.sh`?

```sh
if [ -z "${NGINX_ENTRYPOINT_QUIET_LOGS:-}" ]; then
    exec 3>&1
else
    exec 3>/dev/null
fi
```

Then, doing `>&3` on all the "logging" `echo`s:

```sh
        echo >&3 "$0: Configuration complete; ready for start up"
```

alexandresalome:
I've sent the pull-request modifying the scripts.

Regarding tests, I've verified default and quiet for static file serving. Did not test error cases.

I've verified also the checksum of every `*.sh` files, and they are all kept identical.

Regarding the documentation, I did not found where and/or how to contribute. Is there a way for me to document this variable?

thresheek:
Thanks @alexandresalome for the PR, I will check it out.

Regarding the docs, you can make a PR on https://github.com/docker-library/docs/tree/master/nginx

alexandresalome:
Documentation pull-request prepared, too.

Thank you for your guidance

[...]

PR #424, opened 2020-06-08
Add NGINX_ENTRYPOINT_QUIET_LOGS environment variable

alexandresalome:
**Related issue**: #421

This pull-request adds an environment variable `NGINX_ENTRYPOINT_QUIET_LOGS` to silence information messages when starting up.

By default, verbosity is identical. When this environment variable is defined, the entry point silence information messages, but keep errror messages (on stderr).

### DIFF
--- a/mainline/alpine-perl/10-listen-on-ipv6-by-default.sh
+++ b/mainline/alpine-perl/10-listen-on-ipv6-by-default.sh
@@ -18,10 +18,10 @@ if [ ! -f "/$DEFAULT_CONF_FILE" ]; then
 fi
 
 # check if the file can be modified, e.g. not on a r/o filesystem
-touch /$DEFAULT_CONF_FILE 2>/dev/null || { echo "$ME: Can not modify /$DEFAULT_CONF_FILE (read-only file system?), exiting"; exit 0; }
+touch /$DEFAULT_CONF_FILE 2>/dev/null || { echo >&3 "$ME: Can not modify /$DEFAULT_CONF_FILE (read-only file system?), exiting"; exit 0; }
 
 # check if the file is already modified, e.g. on a container restart
-grep -q "listen  \[::]\:80;" /$DEFAULT_CONF_FILE && { echo "$ME: IPv6 listen already enabled, exiting"; exit 0; }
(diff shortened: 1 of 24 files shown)


mainline/alpine-perl/10-listen-on-ipv6-by-default.sh
```diff
@@ -18,10 +18,10 @@ if [ ! -f "/$DEFAULT_CONF_FILE" ]; then
 fi
 
 # check if the file can be modified, e.g. not on a r/o filesystem
-touch /$DEFAULT_CONF_FILE 2>/dev/null || { echo "$ME: Can not modify /$DEFAULT_CONF_FILE (read-only file system?), exiting"; exit 0; }
+touch /$DEFAULT_CONF_FILE 2>/dev/null || { echo >&3 "$ME: Can not modify /$DEFAULT_CONF_FILE (read-only file system?), exiting"; exit 0; }
```
thresheek, line 21: I'm a bit puzzled why this line has >&3 while it's an error message just like previous echo.

mainline/alpine-perl/10-listen-on-ipv6-by-default.sh
```diff
@@ -18,10 +18,10 @@ if [ ! -f "/$DEFAULT_CONF_FILE" ]; then
...
 
 # check if the file is already modified, e.g. on a container restart
-grep -q "listen  \[::]\:80;" /$DEFAULT_CONF_FILE && { echo "$ME: IPv6 listen already enabled, exiting"; exit 0; }
+grep -q "listen  \[::]\:80;" /$DEFAULT_CONF_FILE && { echo >&3 "$ME: IPv6 listen already enabled, exiting"; exit 0; }
```
thresheek, line 24: Same here, it's an error but is silenced. But not e.g. line 29?

mainline/alpine-perl/10-listen-on-ipv6-by-default.sh
alexandresalome: I changed everything from stdout to `&3`, basically.

Here, if it's an error message, should I redirect it to stderr instead of stdout?

thresheek: OK, got it. Need more coffee.

I'll add 1>&2 to this line, because it should be visible even if logs are silenced.

Thanks!

PR #432, opened 2020-06-25
Moving to alpine:3.12 should fix CVE-2020-14155

prashant4199:


thresheek:
There are no nginx.org-provided packages for alpine 3.12 yet, and this diff does not change the other images that suffer from the same CVE.

Thanks, but no.

J0WI:
Fixed in https://github.com/nginxinc/docker-nginx/commit/aa41ddeef871b7f0ea64a44f26d3f4aa0e6d5e7b

PR #448, opened 2020-08-31
Fix typo in 10-listen-on-ipv6-by-default.sh

gurdiga:
Just that. 🤷🏻‍♂️


thresheek:
I'm not a native english speaker, but to me that's not a typo.  Can anyone weigh in?

tianon:
I think the existing text in the repo was intended to be possessive (ie, should be `package's version`).

Either `packaged` or `package's` seems correct to me 😅

adampweb:
@thresheek That's a typo, because "package" mean "pack" in plurar, "packaged" in turn synonymous with "packed", in singular write 

I'm not a native english speaker too.

thresheek:
Fixed in https://github.com/nginxinc/docker-nginx/commit/a4208f853fa3fcb73a146a268fb2d5f0eee7733b, thank your for the contribution!

PR #457, opened 2020-09-25
Use SIGQUIT instead of SIGTERM for graceful shutdown of nginx

### LINKED ISSUE #377: Use `STOPSIGNAL SIGQUIT` instead of `SIGTERM`
Docker (and, therefore, Kubernetes) send SIGTERM as the default signal to stop containers. Docker will then wait 10 seconds before sending SIGKILL, Kubernetes will wait 30 seconds. They expect containers to exit gracefully within this grace period:

> When you issue a docker stop command Docker will first ask nicely for the process to stop and if it doesn't comply within 10 seconds it will forcibly kill it. If you've ever issued a docker stop and had to wait 10 seconds for the command to return you've seen this in action

From: https://www.ctl.io/developers/blog/post/gracefully-stopping-docker-containers/

> It’s important that your application handle termination gracefully so that there is minimal impact on the end user and the time-to-recovery is as fast as possible!
> 
> In practice, this means your application needs to handle the SIGTERM message and begin shutting down when it receives it. This means saving all data that needs to be saved, closing down network connections, finishing any work that is left, and other similar tasks.

From: https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-terminating-with-grace

However, this is not what nginx does:

> TERM, INT   fast shutdown
> QUIT   graceful shutdown

From: http://nginx.org/en/docs/control.html

The `Dockerfile`s in this project [use](https://github.com/nginxinc/docker-nginx/blob/master/stable/buster/Dockerfile#L101) `STOPSIGNAL SIGTERM` (which actually isn't necessary since this is the default).

What this means in practice is that when Docker or Kubernetes decide to stop a container, nginx will die immediately, closing any connections straight away:

``` bash
$ curl localhost
curl: (52) Empty reply from server
```

It would be much healthier, and comply with the ethos of both Docker and Kubernetes if instead the nginx image made an effort to close all connections before shutting down. This can be achieved nicely with:

``` dockerfile
STOPSIGNAL SIGQUIT
```

I've verified this behaviour locally with [this Dockerfile](https://gist.github.com/nottrobin/79087f07efab61f013c6e7e519a96ad8).

pscheid92:
Resolves #377 

Replace stopsignal SIGTERM with SIGQUIT to ensure a graceful shutdown.
Tested each changed Dockerfile using the [Dockerfile](https://gist.github.com/nottrobin/79087f07efab61f013c6e7e519a96ad8) provided by @nottrobin in #337.

### DIFF
--- a/mainline/alpine-perl/Dockerfile
+++ b/mainline/alpine-perl/Dockerfile
@@ -120,6 +120,6 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 
 EXPOSE 80
 
-STOPSIGNAL SIGTERM
+STOPSIGNAL SIGQUIT
 
 CMD ["nginx", "-g", "daemon off;"]
--- a/mainline/alpine/Dockerfile
+++ b/mainline/alpine/Dockerfile
@@ -119,6 +119,6 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 
 EXPOSE 80
 
-STOPSIGNAL SIGTERM
+STOPSIGNAL SIGQUIT
 
 CMD ["nginx", "-g", "daemon off;"]
--- a/mainline/debian-perl/Dockerfile
+++ b/mainline/debian-perl/Dockerfile
@@ -111,6 +111,6 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 
 EXPOSE 80
 
-STOPSIGNAL SIGTERM
+STOPSIGNAL SIGQUIT
 
 CMD ["nginx", "-g", "daemon off;"]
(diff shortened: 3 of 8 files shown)


taylorb-syd:
This has been discussed a few times, however it was vetoed due to https://trac.nginx.org/nginx/ticket/753. This issue has been fixed in `nginx` since version 1.19.1, therefore there should be no harm in introducing this change for versions later than 1.19.1.

thresheek:
I have merged this change in https://github.com/nginxinc/docker-nginx/commit/3fb70ddd7094c1fdd50cc83d432643dc10ab6243, thanks!

PR #492, opened 2021-01-12
Update alpine to 3.12 for stable images

unreturned:


thresheek:
Thanks, but we'd rather bump those to 3.13 when a new nginx-stable release is out.  No ETA though, and I'd like not to move same stable tag from one OS to another.

Issue #529, opened 2021-04-08
envsubst conflicts with internal variables


Roffild:
envsubst is a very bad solution because variables in Nginx also start with $
```
    set $MAGE_ROOT ${MAGENTO_ROOT}; # Variable: MAGENTO_ROOT
    set $MAGE_MODE ${MAGENTO_RUN_MODE}; # Variable: MAGENTO_RUN_MODE
    set $MFTF_UTILS ${MFTF_UTILS}; # Variable: MFTF_UTILS
    set $WITH_XDEBUG ${WITH_XDEBUG}; # Variable: WITH_XDEBUG

    set $my_fastcgi_pass "fastcgi_backend";
    if ($cookie_XDEBUG_SESSION) {
        set $my_fastcgi_pass "fastcgi_backend_xdebug";
    }

    # Support for SSL termination.
    set $my_http "http";
    set $my_ssl "off";
    set $my_port "80";
    if ($http_x_forwarded_proto = "https") {
        set $my_http "https";
        set $my_ssl "on";
        set $my_port "443";
    }
```

Roffild:
You can create a script on Perl to replace between `%var&` to avoid conflicts.

Why is the dead Perl included in the image, but not Python?

thresheek:
Hi @Roffild, can you explain a bit more about how `envsubst` conflicts with internal configuratoin variables? `envsubst` will not touch anything you are not defining in your docker environment, so it's up to you to configure what's best.

In your case I believe the correct way to use it is to rename the vars passed from environment like `MFTF_UTILS` and `WITH_XDEBUG`, since `envsubst` will replace both entries.

Also, `perl` is included in a Debian-based image since it is a part of a base system, and we cant easily get rid of it.  If you wish to use an image without perl, consider switching to `nginx:alpine` Alpine based-variants.

Roffild:
envsubst simply replaces one text with another without parsing the syntax.

The replacement pattern is the same as the syntax for variables. The chances of replacing the wrong thing and not getting an error is huge. This is a security hole! (Vulnerability)

The replacement pattern MUST NOT be part of the config syntax.

thresheek:
I don't think I agree that it's a vulnerability nor a security hole.

First of all, no replacement is done by default.  You're free to provide your static configuration and it will not be touched by `envsubst`.

Second, `envsubst` won't replace variables it doesn't know of as you have to manually specify them in the environment.

[...]

PR #533, opened 2021-04-10
New template format

Roffild:
Old templates in `/etc/nginx/templates/`
Format `$VAR`

New templates in `/etc/nginx/config/`
Format `^VAR^`

The symbol `^` is selected because it is used only at the beginning of a regular expression and does not appear in the format `^word^`.

The old templates are **deprecated** and dangerous to use. Details at #529

### DIFF
--- a/entrypoint/21-templates.sh
+++ b/entrypoint/21-templates.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+set -e
+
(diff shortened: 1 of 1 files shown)


thresheek:
Sorry, but no.  I don't think having multiple ways achieving the same thing are OK to have in this repo.

PR #549, opened 2021-05-22
Remove non-alphanumeric characters from module name when installing it.

rodrigoaguilera:
I had a couple of problems while building nginx-upload-progress-module
https://github.com/masterzen/nginx-upload-progress-module

First the build was refusing to build the `-dbg` version. Renaming my directory from `nginx-upload-progress-module` to `ngx_http_uploadprogress_module` solved. Added a some a line to docs to clarify that.

Then in the final step tries to install `nginx-module-ngx_http_uploadprogress_module-1.19.10*.apk` but the file generated is
`nginx-module-ngxhttpuploadprogressmodule-1.19.10-r1.apk`

Added a cleanup of the module name to circumvent that.

### DIFF
--- a/modules/Dockerfile.alpine
+++ b/modules/Dockerfile.alpine
@@ -62,6 +62,6 @@ ARG ENABLED_MODULES
 COPY --from=builder /tmp/packages /tmp/packages
 RUN set -ex \
     && for module in $ENABLED_MODULES; do \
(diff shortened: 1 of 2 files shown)


thresheek:
Thanks for bringing this to my attention @rodrigoaguilera !

I've came up with a proper fix in https://github.com/nginxinc/docker-nginx/pull/549/commits/e35a2879e82ef15334320417dbf65df54cd46fd4

PR #560, opened 2021-07-02
Better ENV variables handling.

dfndr-1:
In case when env contain variables with newlines, the variables
substitution script fails, trying to use non-existent variables. Change
env mode to delim different env vars in output with NULL-byte instead of
newline.

### DIFF
--- a/entrypoint/20-envsubst-on-templates.sh
+++ b/entrypoint/20-envsubst-on-templates.sh
@@ -10,7 +10,7 @@ auto_envsubst() {
   local output_dir="${NGINX_ENVSUBST_OUTPUT_DIR:-/etc/nginx/conf.d}"
 
   local template defined_envs relative_path output_path subdir
-  defined_envs=$(printf '${%s} ' $(env -0 | cut -z -d= -f1 | tr -d "\0"))
+  defined_envs=$(cat /proc/self/environ | xargs -I{}  -0 -n1 sh -c 'echo "{}" | sed -n "1 s/^\([^=]*\)=.*$/\${\1} /p"');
   [ -d "$template_dir" ] || return 0
   if [ ! -w "$output_dir" ]; then
     echo >&3 "$ME: ERROR: $template_dir exists, but $output_dir is not writable"
--- a/entrypoint/20-envsubst-on-templates.sh
+++ b/entrypoint/20-envsubst-on-templates.sh
@@ -10,7 +10,7 @@ auto_envsubst() {
   local output_dir="${NGINX_ENVSUBST_OUTPUT_DIR:-/etc/nginx/conf.d}"
 
   local template defined_envs relative_path output_path subdir
-  defined_envs=$(printf '${%s} ' $(env | cut -d= -f1))
+  defined_envs=$(printf '${%s} ' $(env -0 | cut -z -d= -f1 | tr -d "\0"))
   [ -d "$template_dir" ] || return 0
   if [ ! -w "$output_dir" ]; then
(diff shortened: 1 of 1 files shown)


entrypoint/20-envsubst-on-templates.sh
```diff
@@ -10,7 +10,7 @@ auto_envsubst() {
   local output_dir="${NGINX_ENVSUBST_OUTPUT_DIR:-/etc/nginx/conf.d}"
 
   local template defined_envs relative_path output_path subdir
-  defined_envs=$(printf '${%s} ' $(env | cut -d= -f1))
+  defined_envs=$(printf '${%s} ' $(env -0 | cut -z -d= -f1 | tr -d "\0"))
```
tianon, line 13: This probably won't work with Alpine's `env` right? :grimacing: :disappointed:

dfndr-1: Hmm, yep, it won't... It's sad... Will try to find solution suitable for alpine as well.

entrypoint/20-envsubst-on-templates.sh
```diff
@@ -10,7 +10,7 @@ auto_envsubst() {
   local output_dir="${NGINX_ENVSUBST_OUTPUT_DIR:-/etc/nginx/conf.d}"
 
   local template defined_envs relative_path output_path subdir
-  defined_envs=$(printf '${%s} ' $(env -0 | cut -z -d= -f1 | tr -d "\0"))
+  defined_envs=$(cat /proc/self/environ | xargs -I{}  -0 -n1 sh -c 'echo "{}" | sed -n "1 s/^\([^=]*\)=.*$/\${\1} /p"');
```
tianon, line 13: With the limitations of BusyBox `grep`, I did a little "code golf" and this is the best I could come up with that seems to work on both Debian and Alpine:

```suggestion
  defined_envs=$(printf '${%s} ' $(xargs -0n1 -a /proc/self/environ sh -c 'echo "$@" | grep -oEm1 "^[^=]+"' --));
```

(Although still obviously depends on whether the maintainers want this improvement -- my goal is helping it be the best it can be if they do want it. :smile:)

thresheek:
Hi @dfndr-1, can you squash the commits please?  I'd rather have one commit in the repo history.

Thanks!

thresheek:
Merged in https://github.com/nginxinc/docker-nginx/commit/0d38d4931427fea915b2db39c1212db369278d5c

karlderkaefer:
@dfndr-1 I could not reproduce the issue with multiline values. Could give an example for issue? https://gist.github.com/karlderkaefer/c5059afd4c4432b0093d0b08d4128e44#file-nginx-envsubst-sh-L5-L12

thresheek:
Looking at busybox 3.14 release notes:
```
      env: implement -0
```

`alpine:3.16`, which we currently use, has busybox 3.15, so we're safe to use `env -0` there.  I'll adapt the entrypoint to use it.  Note that using xargs and a shell spawned for each variable seems to use a lot of CPU unneccessary, see https://github.com/nginxinc/docker-nginx-unprivileged/issues/107 for more details.

thresheek:
Or maybe we don't even need env | cut here, and can do it with awk

thresheek:
I think something like that would do, but I'm going to do some more testing tomorrow:
```
defined_envs=$(printf '${%s} ' $(awk "END { for (name in ENVIRON) { print ( name ~ /${filter}/ ) ? name : \"\" } }" < /dev/null ))
```

PR #573, opened 2021-09-02
Prevent PATH variable erasing

cartonalexandre:


thresheek:
Out of curiousity, where does this manifest as a problem?

cartonalexandre:
Hi,

I update it as same as line 13, in order to prevent erasing or emptying "PATH" environment variable while starting container. In such case, command find will not be found.
Other utilities like "echo" don't require that, because it's a bash builtins, not based on "PATH" variable.

thresheek:
Maybe it would be easier to set PATH in the file like we do in e.g. `30-tune-worker-processes.sh` ?

cartonalexandre:
Yes you right, just commit the change, I don't manage to stash them..

yosifkit:
If users are starting a container with an empty or changed `PATH` variable (e.g. `docker run -e PATH=""`), then they should expect breakage. `Docker` provides the default `PATH` and so it is reasonable for image creators to rely on that and so it should not be on every image maintainer to protect users from an incorrect `PATH` environment variable.

cartonalexandre:
Thanks for this information, didn't know. I can close this PR.

Issue #561, opened 2021-07-09
The `1.21.1-alpine` tag on Docker Hub is missing multple OS/ARCH entries


JonasAlfredsson:
Hi, 

I am building an image for multiple architectures from this image and ran into the following error for the `linux/386` platform when building from the `1.21.1-alpine` tag:

```
error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: no match for platform in manifest
```

Looking at it on [Docker Hub](https://hub.docker.com/layers/nginx/library/nginx/1.21.1-alpine/images/sha256-833dc94560d9cdb945a0b83bb02b93372ce2dcdf34f4df30fe8f5656ce5d3fb5?context=explore) we indeed see that it is only available for the following architectures:

- linux/arm64/v8
- linux/arm/v7
- linux/arm/v6
- linux/amd64

This is very weird since the [1.21-alpine tag](https://hub.docker.com/layers/nginx/library/nginx/1.21-alpine/images/sha256-c35699d53f03ff9024ce2c8f6730567f183a15cc27b24453c5d90f0e7542daea?context=explore) has also the following ones:

- linux/ppc64le
- linux/s390x
- linux/386

so something must have gone wring in the push that was made.

thresheek:
The x86 issue is known as there is no `mercurial` anymore in alpine 3.14 for this architecture.  Maybe we should do something about it, but I'd rather wait for alpine to sort it out on their own.

For ppc64le and s390x it seems Docker Library builders are stuck/unavailable, so there isnt much we can do at this moment.

cc @tianon @yosifkit

JonasAlfredsson:
Thanks for the swift response.
Glad that this is a known issue, but I am a little bit unsure what this means. Will Alpine stop serving x86 builds of Nginx and this ARCH should be dropped from my automated builds?

yosifkit:
> make: mkdir: Operation not permitted

`s390x` and `ppc64le` are likely to be suffering the same bug as our `php` image builds https://github.com/docker-library/php/issues/1178#issuecomment-875125309.

tianon:
This is fixed for ppc64le and s390x -- for i386, I'm not convinced it's going to be, but maybe? :shrug: :sweat_smile:

JonasAlfredsson:
That is some good news, thanks for posting the update :)
However, what are the plans for 386 going forward? Should this issue remain open, or will that arch be discontinued for Alpine?

[...]

PR #596, opened 2021-11-11
Remove i386 from packages built by upstream for Debian

lucacome:
`i386` packages are not built anymore for `bullseye` https://github.com/nginxinc/docker-nginx/issues/561#issuecomment-965771601


thresheek:
Merged, thanks!

PR #614, opened 2021-12-29
Update nginx mainline to 1.21.5, njs to 0.7.1

lucacome:
Updates nginx and njs to the latest version.


lucacome:
@thresheek looks like the tag for [1.21.5-1](https://hg.nginx.org/pkg-oss/rev/1.21.5-1) is missing

thresheek:
Thanks Luca!

Indeed the tag is not there yet - I'm waiting for some tests to finish today and then will tag it.

PR #653, opened 2022-04-26
Update update.sh

WillemHH-BMW:
Contains a fix for a security vulnerability.
[main/zlib: upgrade to 1.2.12 (CVE-2018-25032)]


thresheek:
No, thanks.  We're using major.minor so the images could be automatically rebuilt by the Library maintainers when there is a new subrelease of an underlying image.

PR #663, opened 2022-05-18
update alpine to 3.15.4 to fix cve

yangyanqing:


thresheek:
No, thanks. We're using major.minor so the images could be automatically rebuilt by the Library maintainers when there is a new subrelease of an underlying image. Refs #653

yangyanqing:
Thanks for your reply.  How offen do you rebuild and push the image ?
On docker hub, the image was still using busybox with cve problem.
We have to clone this repo and rebuild it local, so want to help others to use fresh image directly. :)

@thresheek

thresheek:
There are no explicit predefined dates - so it typically happens whenever a new alpine minor release is out.

PR #666, opened 2022-05-24
Regenerate Alpine Dockerfiles based on latest templates

jnoordsij:
In d07b71be964b53184d829dfd00e2eaa51e4e6ecf a build dependency was changed, but this change was not applied to the stable Dockerfiles. 

Running `update.sh` on the current `master` branch yields the changes in this PR.


thresheek:
That's because stable nginx version (1.20.2) does not support building with pcre2 yet.

jnoordsij:
Ah I see, I was just a little confused when running the script and discovering it caused changes. I'll close this then.

Issue #431, opened 2020-06-24
Environment variable substitution and other entrypoint scripts


rgov:
I have an entry point script that derives a path and then uses `sed` to write it into my nginx config.

I recently moved this over to a script in `/docker-entrypoint.d`. Then I noticed that there is the `envsubst` script, which maybe I could re-use instead of `sed`.

However, there is no way for my earlier entry point script to set an environment variable that gets picked up by the `envsubst` script.

Any thoughts on how this could be accomplished? For instance, I could append to a .env file that gets `source`'d. I'm also willing to concede it might be overly complicated to try to make such a thing work.

tianon:
Why not just `export` in your script before you invoke the image entrypoint?

```bash
#!/bin/sh
export FOO_BAR='baz buzz'
exec /docker-entrypoint.sh "$@"
```

rgov:
Yes I could do this. Thanks for the suggestion. It does push my script outside of the `.d` directory but it's not a solution I would completely reject.

MShekow:
I would like to suggest a different alternative where https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/docker-entrypoint.sh#L22 `source`s the scripts, instead of just executing them. See https://github.com/nginxinc/docker-nginx-unprivileged/issues/86 for details. Would this be something you would consider?

thresheek:
See https://github.com/nginxinc/docker-nginx/pull/687 for another PR willing to do similar things.

I'm reluctant to changing the docker-entrypoint.sh script to source the scripts since it might break someone's workflow.

Adding an option to source a `variables.env` or a similarly named file from a `.d` does look like a good enough option to provide that functionality and make everyone a bit happier.

What do you think?

JuniorJPDJ:
My proposal is adding additional case for eg. `*.envsh` which would be sourced instead of being called directly. This would preserve backwards compatibility and add this feature we want and make some things possible without override of docker container entrypoint.

What do you think guys?

thresheek:
FWIW, my variant of patch is something like:
```
diff --git a/entrypoint/docker-entrypoint.sh b/entrypoint/docker-entrypoint.sh
index 72d5cd9..07f7c93 100755
--- a/entrypoint/docker-entrypoint.sh
+++ b/entrypoint/docker-entrypoint.sh
@@ -13,6 +13,11 @@ if [ "$1" = "nginx" -o "$1" = "nginx-debug" ]; then
[...]

PR #687, opened 2022-07-19
Additional sourced entrypoint files

JuniorJPDJ:
This allows reusing variables declared between entrypoint scripts.

You can create eg. entrypoint starting with 00 to declare some dynamic variables and then use those in envsubst predefined entrypoint.

### DIFF
--- a/entrypoint/docker-entrypoint.sh
+++ b/entrypoint/docker-entrypoint.sh
@@ -16,6 +16,15 @@ if [ "$1" = "nginx" -o "$1" = "nginx-debug" ]; then
         echo >&3 "$0: Looking for shell scripts in /docker-entrypoint.d/"
         find "/docker-entrypoint.d/" -follow -type f -print | sort -V | while read -r f; do
             case "$f" in
+                *.envsh)
+                    if [ -x "$f" ]; then
+                        echo >&3 "$0: Sourcing $f";
+                        source "$f"
+                    else
+                        # warn on shell scripts without exec bit
+                        echo >&3 "$0: Ignoring $f, not executable";
+                    fi
+                    ;;
                 *.sh)
                     if [ -x "$f" ]; then
                         echo >&3 "$0: Launching $f";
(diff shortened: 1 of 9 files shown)


thresheek:
See https://github.com/nginxinc/docker-nginx/issues/431 for some discussion on that matter.

JuniorJPDJ:
Hmm.. you are right. I can add separate case for eg. `.envsh` and that would be sourced. Order will be preserved, compatibility will be preserved..

thresheek:
I actually have that patch already done, just not published. I would like some input on the PR i linked first if that would work for their use cases as well. (I don't see why not, but who knows).

JuniorJPDJ:
I cannot see any PR linked here nor in issue you linked here. You mean you want my feedback in the issue? Sure!

JuniorJPDJ:
Hello. Is there any progress on merging this or providing similar solution? This is blocking for my other changes in funkwhale project.

thresheek:
Thanks for the PR, implementation and reminders!

I've manually adapted the patch and pushed it in https://github.com/nginxinc/docker-nginx/commit/667512852e35b485cc557412e58648cf70a77759

l1553k:
Thanks for this awesome feature! @thresheek can you tell when will it be released as 1.24?

thresheek:
Hi @l1553k! It will be available with the next 1.23 release, however we don't have a strict ETA for it yet. Hopefully in two or three weeks.

blubberblah:
> Thanks for the PR, implementation and reminders!
> 
> I've manually adapted the patch and pushed it in [6675128](https://github.com/nginxinc/docker-nginx/commit/667512852e35b485cc557412e58648cf70a77759)

@thresheek I've patched the docker-entrypoint.sh in a local setup with the current 1.23.1 release. I've noticed that `source` isn't functioning with `/bin/sh` but, it works if I change the shell to `/bin/bash`. Not sure if this can be an issue with the next build.

thresheek:
Oh indeed, it seems debian's dash doesnt support 'source'. Thanks!

thresheek:
Fixed in https://github.com/nginxinc/docker-nginx/commit/cc7ca22aee900542c05594d7a2b18623d84c7d12 - thanks for the notification @blubberblah !

PR #701, opened 2022-09-09
docker-entrypoint: 30-tune-worker-processes.sh: cgroups2 workaround

rockdrilla:
cgroups2 may be set in weird way:

```
$ podman run --rm --network=host debian:stable-slim sh -c 'grep cgroup /proc/self/mountinfo ; echo ; cat /proc/self/cgroup'
754 752 0:27 /../../../../../.. /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime - cgroup2 cgroup2 rw,nsdelegate,memory_recursiveprot

0::/
```

add workaround for that case

Signed-off-by: Konstantin Demin <rockdrilla@gmail.com>

### DIFF
--- a/entrypoint/30-tune-worker-processes.sh
+++ b/entrypoint/30-tune-worker-processes.sh
@@ -158,7 +158,7 @@ __EOF__
     "/")
       foundroot="${found##* }$mountpoint"
       ;;
-    "$mountpoint")
+    "$mountpoint" | /../*)
       foundroot="${found##* }"
       ;;
   esac


rockdrilla:
marking MR as WIP due to discovering one more "broken" case.

thresheek:
Wow, that's crazy.  I wonder if that's documented somewhere in crun or...

In any case, maybe sanitizing `${found%% *}` using realpath(1) would be a better idea?

thresheek:
Yeah, makes sense.  I don't really like the ambiguity in having to specify variants in the case switch, but oh well.

PR #703, opened 2022-09-11
docker-entrypoint: support socket activation

### LINKED ISSUE #702: Don't close file descriptor 3 in entrypoint.sh as it may be used for systemd socket activation
Systemd socket activation does not work because file descriptor 3 is closed in _entrypoint.sh_

The file descriptor 3 is closed here in the  source code

https://github.com/nginxinc/docker-nginx/blob/667512852e35b485cc557412e58648cf70a77759/entrypoint/docker-entrypoint.sh#L6-L10

When a socket-activated systemd service uses one socket, systemd launches the service so that the command gets the activated socket as fd 3 and the environment variable `LISTEN_FDS` set to the value `1`.

When a socket-activated systemd service uses two sockets, systemd launches the service so that the command gets the activated socket as fd 3 and fd 4 and the environment variable `LISTEN_FDS` set to the value `2`.

With more sockets it follows the same scheme as above.

(systemd also sets the environment variable `LISTEN_FDNAMES` but that's not relevant for this GitHub issue) 

Rationale for fixing this issue:

* By not closing file descriptor 3 in _entrypoint.sh_, it would be possible to run _docker.io/library/nginx_ in a socket-activated systemd service with Podman.

* Running  _docker.io/library/nginx_ with  rootless Podman  and socket activation comes with security advantages. It is then possible to run a web site and at the same time restrict internet access for `runc`, `podman` and `conmon`. Only the communication over the activated socket is allowed (see https://github.com/eriksjolund/podman-nginx-socket-activation#possibility-to-restrict-the-network-in-the-container-podman-and-oci-runtime).
* Native network performance when running rootless Podman (see https://github.com/eriksjolund/podman-nginx-socket-activation#native-network-performance-over-the-socket-activated-socket)
* The peer IP address in connections is preserved when running rootless Podman (see https://github.com/eriksjolund/podman-nginx-socket-activation#the-source-ip-address-is-preserved)

Podman supports [socket activation](https://github.com/containers/podman/blob/main/docs/tutorials/socket_activation.md) since 3.4.0. 

Maybe  `systemd-nspawn` could also benefit from this?
I'm not familiar with the command `systemd-nspawn`, but I would assume that it is somehow possible to run _docker.io/library/nginx_ with `systemd-nspawn` and socket activation.

eriksjolund:
Although there is an open feature request to Nginx ("Add optional systemd socket activation support"
https://trac.nginx.org/nginx/ticket/237),
it's possible to use socket activation
by using the environment variable `NGINX`.

When systemd runs a socket-activated service,
systemd sets the environment variables LISTEN_FDS and LISTEN_FDNAMES for the launched process.
In docker-entrypoint.sh set NGINX from a translated LISTEN_FDS value.

For example:

LISTEN_FDS=1
translates to
NGINX=3;

LISTEN_FDS=2
translates to
NGINX=3;4;

LISTEN_FDS=3
translates to
NGINX=3;4;5;

Don't close the file descriptor 3 in the logging machinery.

Fixes #702

Signed-off-by: Erik Sjölund <erik.sjolund@gmail.com>

### DIFF
--- a/entrypoint/docker-entrypoint.sh
+++ b/entrypoint/docker-entrypoint.sh
@@ -3,44 +3,44 @@
 
 set -e
 
-if [ -z "${NGINX_ENTRYPOINT_QUIET_LOGS:-}" ]; then
-    exec 3>&1
-else
-    exec 3>/dev/null
-fi
+entrypoint_log() {
+    if [ -z "${NGINX_ENTRYPOINT_QUIET_LOGS:-}" ]; then
+        echo "$@"
+    fi
+}
 
 if [ "$1" = "nginx" -o "$1" = "nginx-debug" ]; then
     if /usr/bin/find "/docker-entrypoint.d/" -mindepth 1 -maxdepth 1 -type f -print -quit 2>/dev/null | read v; then
-        echo >&3 "$0: /docker-entrypoint.d/ is not empty, will attempt to perform configuration"
+        entrypoint_log "$0: /docker-entrypoint.d/ is not empty, will attempt to perform configuration"
 
-        echo >&3 "$0: Looking for shell scripts in /docker-entrypoint.d/"
+        entrypoint_log "$0: Looking for shell scripts in /docker-entrypoint.d/"
         find "/docker-entrypoint.d/" -follow -type f -print | sort -V | while read -r f; do
             case "$f" in
                 *.envsh)
                     if [ -x "$f" ]; then
-                        echo >&3 "$0: Sourcing $f";
+                        entrypoint_log "$0: Sourcing $f";
                         source "$f"
                     else
                         # warn on shell scripts without exec bit
-                        echo >&3 "$0: Ignoring $f, not executable";
+                        entrypoint_log "$0: Ignoring $f, not executable";
                     fi
                     ;;
                 *.sh)
                     if [ -x "$f" ]; then
-                        echo >&3 "$0: Launching $f";
+                        entrypoint_log "$0: Launching $f";
                         "$f"
                     else
                         # warn on shell scripts without exec bit
-                        echo >&3 "$0: Ignoring $f, not executable";
+                        entrypoint_log "$0: Ignoring $f, not executable";
                     fi
                     ;;
-                *) echo >&3 "$0: Ignoring $f";;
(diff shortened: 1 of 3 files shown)


eriksjolund:
I guess there are different ways how to fix 

* #702

This PR wraps the logging into a separate function where the presence of `NGINX_ENTRYPOINT_QUIET_LOGS` is checked everytime there is log messages to be printed.

Another approach could be to choose another file descriptor than 3 for the logging. If the file descriptor is not used beforehand, then there would be no need to close it.

I've only tried out this PR with one socket (i.e., `LISTEN_FDS=1`) so far.

entrypoint/docker-entrypoint.sh
```diff
@@ -3,44 +3,52 @@
 
 set -e
 
-if [ -z "${NGINX_ENTRYPOINT_QUIET_LOGS:-}" ]; then
-    exec 3>&1
-else
-    exec 3>/dev/null
+. /entrypoint_common.sh
+
+if [ -n "${LISTEN_FDS:-}" ]; then
```
thresheek, line 8: I don't think that (lines 8-17) belongs in this PR or in entrypoint scripts we ship generally.  It seems to  best suited to be a per-solution .envsh script, which you will ship locally with your configuration.

entrypoint/entrypoint_common.sh
```diff
@@ -0,0 +1,5 @@
+entrypoint_log() {
```
thresheek, line 1: While I understand the idea to split common part into a separate file this brings a significant drawback - it will be logged by docker-entrypoint.sh for every image user as a non-executable script.

Code duplication isnt great either, but I think it's still better to make it a bit harder to maintain for me going forward, rather than putting more things in the logs for the users worldwide.

That said, please include this function in every file where we use it.

entrypoint/docker-entrypoint.sh
eriksjolund: Good point. I removed the lines.

entrypoint/entrypoint_common.sh
eriksjolund: Fixed

PR #708, opened 2022-09-28
env file for 20-envsubst-on-templates.sh

Alexey-Tsarev:
Problem: there is no way to provide additional variables to `20-envsubst-on-templates.sh` script.
For example I have some `01-script.sh` which calculates some date and wants to export a result somehow.
A real case, - I faced with the following problem: I need to get FQDN from hostname and put result in a new variable.
Simple `export` does not work.

So I implemented that elegant way to add variables in runtime using separate file `.env`.


thresheek:
Hi @Alexey-Tsarev, it actually is possible to do that, by using *.envsh scripts since https://github.com/nginxinc/docker-nginx/commit/667512852e35b485cc557412e58648cf70a77759

It's not yet in the released versions of the image though.

thresheek:
Closing since a similar PR is already implemented.

PR #714, opened 2022-10-26
Replace envsubst with Perl script

LeviPesin:
Replace envsubst usage with a Perl script based on https://unix.stackexchange.com/a/294836 to remove the need of installing `gettext` and using `envsubst`.

Would be great if somebody knowing Perl check the script (I don't know Perl very well)...


thresheek:
Alpine images don't ship perl so this won't be usable there.

LeviPesin:
I think they do? Docs of Alpine Linux state that Perl is there and [Perl works in its online implementation](https://bellard.org/jslinux/vm.html?url=alpine-x86.cfg&mem=192).

thresheek:
You can try it yourself: `docker run nginx:alpine perl`

LeviPesin:
Closing, then...

PR #717, opened 2022-11-07
Allow extending environment in entrypoints

rafis:
I need to create a custom docker image based on **nginx**. I have created a script `docker-entrypoints.d/00-prepare.sh`:
```sh
#!/bin/sh

if [ "$ENABLE_SOME_LOGIC" == "true" ]; then
   export SOME_VARIABLE="some_value"
fi
```
so that it executes before `20-envsubst-on-templates.sh` and I can use that variables in conf templates.

To my surprise I've found that the `export` doesn't work.


thresheek:
Hi @rafis !

We've recently added a way to source entrypoint scripts, via naming them as `*.envsh` (https://github.com/nginxinc/docker-nginx/commit/667512852e35b485cc557412e58648cf70a77759) - can you please test that and let me know if that works for you?

rafis:
Hi, thank you very much!

PR #667, opened 2022-05-24
Update mainline to Alpine 3.16

jnoordsij:
Alpine 3.16 was [released](https://alpinelinux.org/posts/Alpine-3.16.0-released.html) yesterday. This PR updates the `mainline` image to use the latest Alpine version.

PR #720, opened 2022-11-23
Update to Alpine 3.17

jnoordsij:
Update mainline to Alpine 3.17 (see https://alpinelinux.org/posts/Alpine-3.17.0-released.html).
Requires upstream package builds to be released first.

See also https://github.com/nginxinc/docker-nginx/pull/667.

### DIFF
--- a/mainline/alpine-perl/Dockerfile
+++ b/mainline/alpine-perl/Dockerfile
@@ -3,7 +3,7 @@
 #
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
-FROM alpine:3.16
+FROM alpine:3.17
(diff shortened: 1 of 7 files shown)


thresheek:
Please don't edit dockerfiles manually.

thresheek:
I also think it would make sense to bump stable as well.

thresheek:
Sure, but I don't plan to update the official images hashes to include that change prior to an actual nginx stable release. That means the users will use whatever is published on docker hub / aws public, based on 3.16.

jnoordsij:
That works as well. I'll add another commit then!

PR #731, opened 2022-12-18
Following symlinks when docker-entrypoint.sh starts.

walkure:
Some k8s implentations (I use k0s v1.25.2+k0s.0) use symlinks when mounting ConfigMap as volume.

image

Following `find` supports symbolic linked contents. But the first `find` does not follow symlinks, it is not processed.

https://github.com/nginxinc/docker-nginx/blob/5ce65c3efd395ee2d82d32670f233140e92dba99/entrypoint/docker-entrypoint.sh#L17

---

Since `find` (L17) includes dotdirs when following symlinks, found shell scripts are evaluated multiple times (at my environment.).
Excluding dotdirs(like below) will resolve it , but I'm afraid of breaking backward compatibility.

```sh
find "/docker-entrypoint.d/" -follow -type d -name '.*' -prune -o -type f -print
```

### DIFF
--- a/entrypoint/docker-entrypoint.sh
+++ b/entrypoint/docker-entrypoint.sh
@@ -10,7 +10,7 @@ entrypoint_log() {
 }
 
 if [ "$1" = "nginx" -o "$1" = "nginx-debug" ]; then
-    if /usr/bin/find "/docker-entrypoint.d/" -mindepth 1 -maxdepth 1 -type f -print -quit 2>/dev/null | read v; then
+    if /usr/bin/find "/docker-entrypoint.d/" -follow -mindepth 1 -maxdepth 1 -type f -print -quit 2>/dev/null | read v; then
         entrypoint_log "$0: /docker-entrypoint.d/ is not empty, will attempt to perform configuration"
 
         entrypoint_log "$0: Looking for shell scripts in /docker-entrypoint.d/"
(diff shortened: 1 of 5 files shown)


profiprog:
I also have issue related whit this. I use image in kubernetes and mount Config into `/docker-entrypoint.d/`. 
Structure looks like:
```
# tree /docker-entrypoint.d/ -a
/docker-entrypoint.d/
├── ..2023_01_16_13_30_24.2292211062
│   └── setup-data.sh
├── ..data -> ..2023_01_16_13_30_24.2292211062
└── setup-data.sh -> ..data/setup-data.sh
```

And command: `/usr/bin/find "/docker-entrypoint.d/" -mindepth 1 -maxdepth 1 -type f -follow -print -quit`
requires option `-follow` to work as expected.

However command: `find "/docker-entrypoint.d/" -follow -type f -print` \
also needs options `-mindepth 1 -maxdepth 1`
```
# find "/docker-entrypoint.d/" -follow -type f -print
/docker-entrypoint.d/..data/setup-data.sh
/docker-entrypoint.d/setup-data.sh
/docker-entrypoint.d/..2023_01_16_13_30_24.2292211062/setup-data.sh

# find "/docker-entrypoint.d/" -follow -mindepth 1 -maxdepth 1 -type f -print
/docker-entrypoint.d/setup-data.sh
```

PR #742, opened 2023-01-13
Update modules Dockerfiles

netsandbox:
Update the modules/Dockerfile:

* use apt-get instead of apt in modules/Dockerfile
  Do not use apt as it is meant to be an end-user tool.

* use "-o pipefail" for RUN commands in modules/Dockerfile
  see:
  https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#using-pipes
  https://github.com/hadolint/hadolint/wiki/DL4006

### DIFF
--- a/modules/Dockerfile
+++ b/modules/Dockerfile
@@ -12,8 +12,8 @@ RUN set -ex \
 COPY ./ /modules/
 
 RUN set -ex \
-    && apt update \
-    && apt install -y --no-install-suggests --no-install-recommends \
+    && apt-get update \
+    && apt-get install -y --no-install-suggests --no-install-recommends \
                 patch make wget mercurial devscripts debhelper dpkg-dev \
                 quilt lsb-release build-essential libxml2-utils xsltproc \
                 equivs git g++ libparse-recdescent-perl \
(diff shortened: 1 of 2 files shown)


thresheek:
Thanks for the contribution @netsandbox !

Any chance you would provide the same fix for pipefail for an alpine-based modules Dockerfile?

netsandbox:
@thresheek I have updated now also the alpine Dockerfile.

modules/Dockerfile
```diff
@@ -3,17 +3,17 @@ FROM ${NGINX_FROM_IMAGE} as builder
 
 ARG ENABLED_MODULES
 
-RUN set -ex \
-    && if [ "$ENABLED_MODULES" = "" ]; then \
+SHELL ["/bin/bash", "-exo", "pipefail", "-c"]
```
thresheek, line 6: Wouldnt this affect all further invocations of a resulting image?  Maybe we should just do it like we did with "set -ex", and add pipefail there for the scripts we explicitely chosen to have it?

netsandbox: What negative effect are you thinking of?
In general you don't want to have a image build where a pipe silently fails.
And if you add it only on demand, then you have to remember this when you maybe add later a pipe command.

Many people use this in Dockerfiles and shell scripts for years (including me), without experiencing any problem with this.
It is also the recommended by hadolint: https://github.com/hadolint/hadolint/wiki/DL4006

netsandbox:
@thresheek anything else I can do to get this merged?

thresheek:
Thanks @netsandbox !

PR #749, opened 2023-02-01
Add OCI labels

lucacome:
Adds labels according to the spec https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

### DIFF
--- a/Dockerfile-alpine-perl.template
+++ b/Dockerfile-alpine-perl.template
@@ -1,5 +1,7 @@
 FROM nginx:%%NGINX_VERSION%%-alpine
 
+LABEL org.opencontainers.image.base.name="docker.io/library/nginx:%%NGINX_VERSION%%-alpine"
+
 RUN set -x \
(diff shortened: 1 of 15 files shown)


thresheek:
Hello @lucacome !

Given the fact that we also mirror those containers to other registries, e.g. AWS ECR Public, how should we deal with org.opencontainers.image.base.name?

lucacome:
@thresheek the base images are always coming from docker.io right?

thresheek:
debian:slim and alpine are indeed coming from docker.io, correct.

tianon:
Related: https://github.com/docker-library/official-images/issues/3540, https://github.com/docker-library/official-images/issues/12580 :eyes:

tianon:
Yeah, I personally avoid `LABEL` entirely, but if you do want to use them, they should be compliant to the recommendations for annotations in the OCI's image-spec.

Issue #422, opened 2020-06-07
Using environment variables in nginx configuration not working


mjswensen:
According to [the documentation](https://hub.docker.com/_/nginx), I should be able to provide a template directory and the container will automatically extract and replace environment variables, dropping the resulting files in `/etc/nginx/conf.d/`. I don't see that behavior.

# Base reproducible case

Given the following files (copied and pasted from the example in the documentation):

`docker-compose.yml`:

```
version: '3.7'

services:
  web:
    image: nginx
    volumes:
      - ./templates:/etc/nginx/templates
    ports:
      - "8080:80"
    environment:
      - NGINX_HOST=foobar.com
      - NGINX_PORT=80
```

`templates/foobar.conf.template`:

```
server {
  listen       ${NGINX_PORT};
}
```

I then run `docker-compose up` and attach a separate shell to my running `web` container.

From within the container I observe the following:

```sh
$ ls /etc/nginx/templates/
foobar.conf.template
# as expected

$ ls /etc/nginx/conf.d/
default.conf
# where is foobar.conf?
```

Thank you for your help and forgive me if I'm missing an obvious step!

TamerShlash:
Same problem here. I also tried manually running `nginx -g 'daemon off;' -T` and it's using the original `default.conf`only.

thresheek:
I've just tried it and it works.

Can you "docker pull nginx" or "nginx:1.19.0-alpine" and then `docker-compouse up -d` and then show the output of `docker logs` for the launched container?

TamerShlash:
For my case, I know what the problem was.

I was overriding `CMD` in my `Dockefile`. By taking a look at [the content](https://github.com/nginxinc/docker-nginx/blob/master/stable/alpine/docker-entrypoint.sh#L6) of the entrypoint script of the Nginx docker image:

```
if [ "$1" = "nginx" -o "$1" = "nginx-debug" ]; then
```

This means that, if your custom `CMD` does not start with `nginx` or `nginx-debug` (which was the case for me), none of the `/docker-entrypoint.d` scripts will be run because of how `docker-entrypoint.sh` works. The Nginx template compilation step is one of those scripts.

This is very unclear in the `README` of the image on dockerhub and I was only able to figure it out by looking at the code of the entrypoint script.

[...]

PR #752, opened 2023-02-08
Run docker-entrypoint.d scripts for any CMD.

Bad-ptr:
Why is that check [`if [ "$1" = "nginx" -o "$1" = "nginx-debug" ]`](https://github.com/nginxinc/docker-nginx/blob/761fffeba0d867d6e80d38998073e0eaa456bb02/entrypoint/docker-entrypoint.sh#L12) needed in `docker-entrypoint.sh`?
I think it will be usefull to run `docker-entrypoint.d/` scripts regardless of arguments passed.
This will allow to specify a custom `CMD` to execute, while still perform `/etc/nginx/templates` substitutions.

Related issue https://github.com/nginxinc/docker-nginx/issues/422 .

### DIFF
--- a/entrypoint/docker-entrypoint.sh
+++ b/entrypoint/docker-entrypoint.sh
@@ -9,39 +9,37 @@ entrypoint_log() {
     fi
 }
 
-if [ "$1" = "nginx" -o "$1" = "nginx-debug" ]; then
-    if /usr/bin/find "/docker-entrypoint.d/" -mindepth 1 -maxdepth 1 -type f -print -quit 2>/dev/null | read v; then
-        entrypoint_log "$0: /docker-entrypoint.d/ is not empty, will attempt to perform configuration"
+if /usr/bin/find "/docker-entrypoint.d/" -mindepth 1 -maxdepth 1 -type f -print -quit 2>/dev/null | read v; then
+    entrypoint_log "$0: /docker-entrypoint.d/ is not empty, will attempt to perform configuration"
 
-        entrypoint_log "$0: Looking for shell scripts in /docker-entrypoint.d/"
-        find "/docker-entrypoint.d/" -follow -type f -print | sort -V | while read -r f; do
-            case "$f" in
-                *.envsh)
-                    if [ -x "$f" ]; then
-                        entrypoint_log "$0: Sourcing $f";
-                        . "$f"
-                    else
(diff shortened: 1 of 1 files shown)


thresheek:
We're basically following the same approach as other official images do, e.g.:
https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L302
https://github.com/docker-library/cassandra/blob/master/docker-entrypoint.sh#L43
https://github.com/docker-library/mongo/blob/master/docker-entrypoint.sh#L245
https://github.com/apache/couchdb-docker/blob/main/3.3.1/docker-entrypoint.sh#L27
https://github.com/Kong/docker-kong/blob/master/docker-entrypoint.sh#L31
https://github.com/MariaDB/mariadb-docker/blob/master/docker-entrypoint.sh#L486
https://github.com/docker-library/mysql/blob/master/template/docker-entrypoint.sh#L386

I'm not sure allowing any CMD is a great idea because we will be breaking the principle of least surprise?

Bad-ptr:
OK. I see, thanks for links.
But how did this become the standard?

Don't you think that it is more logical to override `ENTRYPOINT` if you want to change default "startup actions" of an image?
With this approach you have two options:
1) use `CMD` if you want default "startup actions".
2) use `ENTRYPOINT` to completely override container startup.

Or am I missing something?

yosifkit:
> But how did this become the standard?

That's partially the fault of official images [consistency](https://github.com/docker-library/official-images/tree/ce5d1cb57568b764206a4a4eec5d16aff834325f#consistency), our [hacky test](https://github.com/docker-library/official-images/tree/ce5d1cb57568b764206a4a4eec5d16aff834325f/test/tests/override-cmd) to ensure it is followed, and us creating the `.d/` script usage for official images. It just made sense for all of them to not do anything related to `postgres`, for example, if `postgres` wasn't the thing being run. You would not want the database started, data inserted, and the database stopped if you only asked for `bash` and happened to have environment variables defined.

We don't technically have a specific rule against the entrypoint script making file changes before executing the command (assuming it doesn't make any output), but from a new user perspective, it could be more surprising that something more than just `bash` happened when they ran `docker run -it --rm nginx bash`.

Bad-ptr:
OK. Thanks for the explanation!

Issue #740, opened 2023-01-09
CVE-2022-3996


vitoyucepi:
## Summary

| Package    | Vulnerability ID | Severity | Installed Version | Fixed Version |
|------------|------------------|----------|-------------------|---------------|
| libcrypto3 | CVE-2022-3996    | HIGH     | 3.0.7-r0          | 3.0.7-r2      |
| libssl3    | CVE-2022-3996    | HIGH     | 3.0.7-r0          | 3.0.7-r2      |

https://gitlab.alpinelinux.org/alpine/aports/-/commit/2bd80c4a148ceea58da585e0f7bb4d60bfdb062b

## Images
- `nginx:mainline-alpine`
- `nginx:mainline-alpine-perl`
- `nginx:mainline-alpine-slim`

## Additional

- The images mentioned earlier will be rebuilt if *alpine* or *nginx* receive an update.
- Use `nginx:stable-alpine`, it uses *openssl1.1*.
- You can update dependencies by yourself.
  ```dockerfile
  FROM nginx:1.23.3-alpine-slim

  ENV OPENSSL_VERSION=3.0.7-r2

  RUN set -eux; \
        apk add --no-cache \
          libssl3="${OPENSSL_VERSION}" \
          libcrypto3="${OPENSSL_VERSION}" \
        ;
  ```
- You can rebuild the image by yourself.
- You can use `alpine:3.17` as your base image and install *nginx* from the repository.
- Official reports
  https://www.openssl.org/news/secadv/20221213.txt
- Based on #738.

### Why I think it's safe to upgrade these dependencies by yourself

1. These versions are ABI compatible.
2. Alpine maintainers didn't rebuild *nginx* in repos.

yosifkit:
This is already in process since the merge of https://github.com/docker-library/official-images/pull/13854. With an `Alpine` base image update, the Alpine-based [`nginx`](https://hub.docker.com/_/nginx) images get automatically rebuilt and pushed to Docker Hub for all supported architectures. (It seems that the `nginx` images might already be done)

```console
$ docker pull nginx:mainline-alpine
mainline-alpine: Pulling from library/nginx
8921db27df28: Pull complete 
83e90619bc2e: Pull complete 
d52adec6f48b: Pull complete 
10eb2ce358fa: Pull complete 
a1be370d6a52: Pull complete 
689b9959905b: Pull complete 
c7a81ce22aac: Pull complete 
[...]

PR #753, opened 2023-02-08
Upgrade openssl-dev package to keep libssl up to date

zapotek6:
The current base image, alpine 3.17, does not bundle with the latest version of libssl and the docker build process does not update the libssl library. So building the image does not guaranteee to have installed the latest version. This change take care to build the image using the latest libssl version available from alpine packages repository.

### DIFF
--- a/mainline/alpine-slim/Dockerfile
+++ b/mainline/alpine-slim/Dockerfile
@@ -21,6 +21,7 @@ RUN set -x \
 # install prerequisites for public key and pkg-oss checks
     && apk add --no-cache --virtual .checksum-deps \
         openssl \
+	openssl-dev \
     && case "$apkArch" in \
         x86_64|aarch64) \
 # arches officially built by upstream


thresheek:
Hi @zapotek6!

I don't think this change is correct.

If you manually rebuild the images, please start with mainline/alpine-slim, and it will pull the latest openssl/libbsl3.  Then, based on that slim image, build the one you're targeting for (mainline/alpine, looking at the patch.)

zapotek6:
Hi @thresheek , thank you to taking time to review my pull request.

I had tried to manually rebuild the nginx mainline alpine-slim, but the libssl was old

> libssl3-3.0.7-r2 x86_64 {openssl} (Apache-2.0) [installed]

The reason of the manual rebuild is that some security fix for openssl has been released, so I decided to start the upgrade process for the production services (haproxy, nginx).

The latest official images of `haproxy:2.7.2-alpine` and `nginx:1.23.3-alpine-slim` does not contain the fixed version of libssl. I started to rebuild them from the official Dockerfile.
For haproxy I ended up with the right libssl version (libssl3-3.0.8-r0) for nginx no.

Both start from the same base image alpine:3.17 so I started digging into.

I ported some changes from the haproxy Dockerfile[1] to the nginx Dockerfile one by one. The `apk add openssl-dev` command is the one that upgraded the libssl.
Reading the apk docs I've understood that the `add` command is used to add/upgrade a package.

From my point of view this is a safe a non destructive approach, because if the base image is not updated in a timely fashion and fixes for libssl are already available we are assuring that the produced image will use the latest available version of the libssl from the alpine package reposity.

I don't know if it is the same thing that happen in the Dockerfile when building an architecture different from x86_64|aarch64.

I'm probably doing something wrong, so bear with me.

thanks for the patience

[1]: https://github.com/docker-library/haproxy/blob/079f57941442130b1f47fbb30181390c8fdfd708/2.7/alpine/Dockerfile

thresheek:
Thanks for such a detailed comment @zapotek6 !

Indeed, I was wrong in my assumption that installing openssl and not openssl-dev will trigger the same logic - but it seems openssl does not depend on a specific libssl version..

In any case, I don't think it's a feasible way for us to stay on top of the things security-wise.  It's up for Alpine image maintainers to provide a secure by default image, which we can build upon.  See e.g. https://github.com/nginxinc/docker-nginx/issues/740#issuecomment-1376467604 for more details.

yosifkit:
Alpine is working on updating more of their base images (https://github.com/docker-library/official-images/pull/14049#issuecomment-1423476110), which will then cause Official Images to rebuild all dependent images (like `nginx:*alpine` images).

PR #767, opened 2023-04-03
Add gpg key to apt instead of keyring

lucacome:
With this change `apt` will automatically load the key, removing the need to specify it in `nginx.list`.


thresheek:
Hi Luca!

I think apt-key (and trusted.gpg.d usage) is deprecated these days: https://manpages.debian.org/bullseye/apt/apt-key.8.en.html

While it will certainly work for the foreseeable future, it's best not to revert to the olden ways...

lucacome:
I didn't realize the folder was deprecated too, sorry for the noise!

PR #776, opened 2023-05-10
Update to Alpine 3.18

wheatevo:
This change updates the mainline and stable Alpine images to use 3.18.0 instead of 3.17.0.

Release post:
https://alpinelinux.org/posts/Alpine-3.18.0-released.html


thresheek:
We don't have nginx.org binaries for Alpine 3.18 yet.  Thanks for a headsup though.

PR #782, opened 2023-05-18
Add an entrypoint for configuring and templating streams

iain-henderson:
Add an entrypoint to allow for templated stream configurations.


thresheek:
Hi @iain-henderson!

Thanks for contribution!

I'm not a huge fan of introducing yet another script to do that - maybe combine with the already existing http-based one?

I think it'd be ok if for stream we iterate over `.stream-template` or similarly named files and append the stream block if stream templates were found/processed?

iain-henderson:
All of that makes sense (I could honestly go either way, but consolidating the templating logic seems like a good choice).

thresheek:
Many thanks!

I've modified the changes a bit and pushed it.

Notable changes to your PR include:
- /etc/nginx/stream-conf.d as a path (corresponds to what https://github.com/nginxinc/kubernetes-ingress/ already uses)
- dropped maxdepth in a search for .stream-template files since we know there can be nested directories in template dirs.
- check for nginx.conf to be writable
- create and respect NGINX_ENVSUBST_STREAM_OUTPUT_DIR

I hope to make it available with the next mainline release.  Let's wait a bit to get it propagated to stable.

Issue #770, opened 2023-04-19
Allow even more advanced template configuration


LeviPesin:
I suggest to allow in templates some kind of conditional (also based on environment variables) execution and loops.

My use-case: I develop a NGINX-based proxy that among other parameters (in environment variables) takes in a list of websocket URLs and a list of backends to which it should proxy them. Allowing to write something like `#for (($url, $backend) in ($URLS.split(), $BACKENDS.split())) {location $url {include websocket.conf; proxy_pass $backend}}` will be extremely useful.

thresheek:
I think that's reinventing a proper templating system and templating language with it, but poorly.  There are tools to generate complex configuration and frankly, I don't think we need to reinvent them just for nginx.

LeviPesin:
> There are tools to generate complex configuration and frankly, I don't think we need to reinvent them just for nginx.

Can you suggest some? Or maybe an entrypoint script could be added to run one on the configuration.

LeviPesin:
So I tried to use them and I think they are just too bloated for this relatively simple task...
I did some search and found [an unmaintained repo](https://github.com/saghul/njk) and [this StackOverflow question](https://stackoverflow.com/questions/7491352/looking-for-a-standalone-command-line-code-generator-script)... From them I think I will use [gomplate](https://docs.gomplate.ca/).

PR #788, opened 2023-05-29
Add gomplate support

LeviPesin:
Related issue: https://github.com/nginxinc/docker-nginx/issues/770

This PR adds an entrypoint script for running [`gomplate`](https://docs.gomplate.ca/) on templates. It is not included in the image by default and requires users to be installed manually (unfortunately it is very large -- more than 37 MB. As an alternative <1 MB [njk](https://github.com/saghul/njk) can be considered, but it seems to be unmaintained for almost two years).

### DIFF
--- a/entrypoint/25-gomplate.sh
+++ b/entrypoint/25-gomplate.sh
@@ -1,0 +1,63 @@
+#!/bin/sh
+
+set -e
+
+ME=$(basename $0)
+
+entrypoint_log() {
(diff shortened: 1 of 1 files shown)


thresheek:
I think adding a code that would be launched and do nothing until you specifically make an image on top of the base one is a bad idea.  First, it will just waste cpu cycles, and second, it will never be tested.

If an idea is intead to provide an example, a documentation is a better choice.

LeviPesin:
I'm fine with adding it to the documentation. Can you please do that?

PR #794, opened 2023-06-07
Fix unbound variable issue in local resolver entrypoint script

euven:
Using the `master` branch here to build an image results in the following error, when trying to run the built image:
```
$ docker run --rm my-image-built-with-master
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: Unsupported distribution
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: 9: /docker-entrypoint.d/15-local-resolvers.envsh: NGINX_ENTRYPOINT_LOCAL_RESOLVERS: parameter not set
```

This pull request fixes the unset parameter issue :)

### DIFF
--- a/entrypoint/15-local-resolvers.envsh
+++ b/entrypoint/15-local-resolvers.envsh
@@ -6,6 +6,6 @@ set -eu
 LC_ALL=C
 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
-if [ "${NGINX_ENTRYPOINT_LOCAL_RESOLVERS}" ]; then
+if [ ! -z "${NGINX_ENTRYPOINT_LOCAL_RESOLVERS+defined}" ]; then
   export NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf)
 fi


thresheek:
Hi @euven, which distribution is that so I could reproduce the issue locally?

Thanks!

euven:
Hi @thresheek - I believe you can test this with debian as it contains the file in question: https://github.com/nginxinc/docker-nginx/tree/master/stable/debian

thresheek:
Thanks.  I've been able to reproduce it.

I believe then the fix is incorect, though, as `-v VARNAME` was added in bash 4.2, and we're on dash in the debian-based image:

```
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: 9: [: -v: unexpected operator
```

Something like that should work though (tested on both Debian and Alpine):

```patch
diff --git a/entrypoint/15-local-resolvers.envsh b/entrypoint/15-local-resolvers.envsh
index 12d9284..7dff10d 100755
--- a/entrypoint/15-local-resolvers.envsh
+++ b/entrypoint/15-local-resolvers.envsh
@@ -6,6 +6,6 @@ set -eu
 LC_ALL=C
 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

-if [ "${NGINX_ENTRYPOINT_LOCAL_RESOLVERS}" ]; then
-  export NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf)
-fi
+[ "${NGINX_ENTRYPOINT_LOCAL_RESOLVERS:-}" ] || continue
+
+export NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf)
```

euven:
Hi @thresheek - apologies, I was testing with bash :facepalm:
Thanks for testing :)

In the fix you suggested, I think `NGINX_LOCAL_RESOLVERS` will always get exported, which is not what we want, right?

Anyways, I've pushed an updated patch - let me know if you have any questions.

thresheek:
Thanks!

> In the fix you suggested, I think NGINX_LOCAL_RESOLVERS will always get exported, which is not what we want, right?

Not really, since `|| continue` will make the script exit if the variable is unset.

thresheek:
Here's an updated patch:
```patch
diff --git a/entrypoint/15-local-resolvers.envsh b/entrypoint/15-local-resolvers.envsh
index 12d9284..9306215 100755
--- a/entrypoint/15-local-resolvers.envsh
+++ b/entrypoint/15-local-resolvers.envsh
@@ -6,6 +6,6 @@ set -eu
 LC_ALL=C
 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

-if [ "${NGINX_ENTRYPOINT_LOCAL_RESOLVERS}" ]; then
-  export NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf)
-fi
+[ "${NGINX_ENTRYPOINT_LOCAL_RESOLVERS:-}" ] || return 0
+
+export NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf)
```

thresheek:
Pushed in 5673d69221fa0367040d60397cc0d39ee892f4d5

PR #797, opened 2023-06-07
feat(Security): Upgrade alpine packages at build time

reneleonhardt:
Upgrade alpine packages at build time, this fixes all current CVEs.

### DIFF
--- a/Dockerfile-alpine-slim.template
+++ b/Dockerfile-alpine-slim.template
@@ -12,6 +12,8 @@ RUN set -x \
     && apkArch="$(cat /etc/apk/arch)" \
     && nginxPackages="%%PACKAGES%%
     " \
+# Update all packages for security reasons
+    && apk --no-cache upgrade \
 # install prerequisites for public key and pkg-oss checks
     && apk add --no-cache --virtual .checksum-deps \
         openssl \


yosifkit:
We recommend against using blanket package upgrades (`apt-get upgrade`/`apk upgrade`/`yum upgrade`/`yum update`) for official-images. When package upgrades are applied in a dependent image, it duplicates content of the base image, making the image larger than necessary. It also only delays the inevitable "there are outdated packages". The Official Images build pipeline makes heavy use of `docker build` cache, so we make periodic base image updates to then fully rebuild all dependent images (e.g. the [Debian image updates](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fdebian)).

>We strive to publish updated images at least monthly for Debian. We also rebuild earlier if there is a critical security need, e.g. [docker-library/official-images#2171](https://github.com/docker-library/official-images/issues/2171). Many Official Images are maintained by the community or their respective upstream projects, like Ubuntu, Alpine, and Oracle Linux, and are subject to their own maintenance schedule. These refreshed base images also means that any other image in the Official Images program that is `FROM` them will also be rebuilt (as described in [the project `README.md` file](https://github.com/docker-library/official-images#library-definition-files)).
>
> \- https://github.com/docker-library/faq/tree/0ad5fd60288109c875a54a37f6581b2deaa836db#why-does-my-security-scanner-show-that-an-image-has-cves

---

Reading issues on Alpine Linux like https://gitlab.alpinelinux.org/alpine/aports/-/issues/14973, I think Alpine releases are imminent (which would mean new Alpine images and all dependent Official Images being rebuilt). They are having trouble with their ARM build infrastructure, so I think it is delayed while they work that out.

PR #818, opened 2023-07-20
Skip configuring image in read-only systems

alessfg:
The image in its current state fails when being run in read-only systems (https://github.com/nginxinc/docker-nginx-unprivileged/issues/43). This PR aims to fix that.

### DIFF
--- a/entrypoint/docker-entrypoint.sh
+++ b/entrypoint/docker-entrypoint.sh
@@ -10,37 +10,44 @@ entrypoint_log() {
 }
 
 if [ "$1" = "nginx" -o "$1" = "nginx-debug" ]; then
-    if /usr/bin/find "/docker-entrypoint.d/" -mindepth 1 -maxdepth 1 -type f -print -quit 2>/dev/null | read v; then
-        entrypoint_log "$0: /docker-entrypoint.d/ is not empty, will attempt to perform configuration"
+    # check if /etc/nginx directory is writable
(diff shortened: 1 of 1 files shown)


yosifkit:
If the user is providing or relying on the scripts in `/docker-entrypoint.d/` but not also providing the place they are expected to write as writeable, then the image shouldn't just ignore the scripts. Failure is the expected result so that they can fix their setup instead of getting a misconfigured service.

alessfg:
The intent (and the overall intent of the thread in the linked issue) of this PR is for the scripts to be ignored in read-only systems where the overall goal is for those scripts to be ignored altogether.

An alternative would be to provide an even more "minimal" version of the images that strips all the entrypoint scripts altogether, but I was trying to avoid adding yet another variant.

yosifkit:
In that case, wouldn't the simplest solution be to just skip the entrypoint altogether since it isn't providing anything except running the default command?

This one fails since it can't create `var/cache/nginx/client_temp` or `/var/run/nginx.pid`, which this PR can't really address:
```console
$ docker run -d --read-only --entrypoint nginx nginx:[version] -g 'daemon off'
```

The only difference between these is a little extra output at the beginning, but both run nginx:
```console
$ docker run -it --rm --read-only --tmpfs /var/cache/nginx/ --tmpfs /var/run/ --entrypoint nginx nginx -g 'daemon off;'
$ docker run -it --rm --read-only --tmpfs /var/cache/nginx/ --tmpfs /var/run/ nginx
```

PR #819, opened 2023-07-20
Correctly check if files are available in Alpine Linux, address some shell linter warnings, and do some minor script beautifying

alessfg:
* Fix small bug with using the `-n` flag instead of the `-f` flag when checking if files are available (https://github.com/nginxinc/docker-nginx/blob/master/Dockerfile-alpine-slim.template#L76-L78)
* Fix some shell linter warnings
* Minor script formatting/beautifying.


thresheek:
Hello!

Can you split actual fixes (like -n vs -f) and unrelated changes into separate commits?

Also, I'm not sure whitespace (and beautyfing, like splitting oneliners into multilines) changes are worth the effort, since by default vim will not honor modelines anyway - and they would pollute the history and blame.

alessfg:
See https://github.com/nginxinc/docker-nginx/pull/824 😄

alessfg:
I do think there is something to be gained in readability for folks that are looking through the scripts either here or using a code editor instead of VIM, but I get your point. I will tweak the PR to get rid of some of the superfluous "beautifying" (e.g. revert the oneliner to multiline changes).

PR #821, opened 2023-07-25
Update gpg dir

vndroid:
- according to [Debian UseThirdParty](https://wiki.debian.org/DebianRepository/UseThirdParty), gpg trust directory changed to `/etc/apt/keyrings`
- Debian 11 (bullseye) need to create dir first, This can be removed in subsequent upgrades to the new version.

> If future updates to the certificate will be managed by an apt/dpkg package as recommended below, then it SHOULD be downloaded into /usr/share/keyrings using the same filename that will be provided by the package. If it will be managed locally , it SHOULD be downloaded into /etc/apt/keyrings instead.

### DIFF
--- a/mainline/debian/Dockerfile
+++ b/mainline/debian/Dockerfile
@@ -19,7 +19,7 @@ RUN set -x \
     && apt-get install --no-install-recommends --no-install-suggests -y gnupg1 ca-certificates \
     && \
     NGINX_GPGKEY=573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62; \
-    NGINX_GPGKEY_PATH=/usr/share/keyrings/nginx-archive-keyring.gpg; \
+    NGINX_GPGKEY_PATH=/etc/apt/keyrings/nginx-archive-keyring.gpg; \
     export GNUPGHOME="$(mktemp -d)"; \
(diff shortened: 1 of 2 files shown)


thresheek:
Hi @vndroid !

I'm not sure Debian wiki counts as an official documentation, however [man 5 sources.list on Debian 12](https://manpages.debian.org/bookworm/apt/sources.list.5.en.html) says:

``
The recommended locations for keyrings are /usr/share/keyrings for keyrings managed by packages, and /etc/apt/keyrings for keyrings managed by the system operator.
``

Looking at other official images, it seems to be all over the place at this moment.  I don't mind this change, but let's wait until we move stable images to Debian 12 so we can drop the mkdir altogether.

vndroid:
> /usr/share/keyrings for keyrings managed by packages

`/usr/share/keyrings for keyrings managed by packages` means the Debian official gpg key is stored in it. Then the others key will be stored in the /etc/ dir.

Example : https://docs.docker.com/engine/install/debian/ Docker has moved into the new directory. But I agree with you.

You can merge it in first, and then wait until a later stage to just remove the create statement and comment information.

thresheek:
thanks @vndroid !

PR #855, opened 2024-01-04
Add IPv6 support for local-resolvers substitution script

serverwentdown:
### Proposed changes

When running in IPv6 networks, the `15-local-resolvers.envsh` script will create invalid nginx configuration because the IPv6 address is not wrapped in brackets as Nginx requires. This change checks for the presence of colons (which all valid IPv6 addresses should have), and wraps the address in square brackets. `/etc/resolv.conf` syntax should never present a host and port together, so it is a valid assumption that an IPv4 nameserver configuration will never have a colon present.

This is incredibly useful for getting DNS working in IPv6 Kubernetes environments.

### Checklist

- [x] If applicable, I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/.test/config.sh
+++ b/.test/config.sh
@@ -4,6 +4,7 @@ imageTests+=(
         static
         templates
         templates-resolver
+        templates-resolver-ipv6
         workers
         modules
 	'
--- a/.test/tests/templates-resolver-ipv6/expected-std-out.txt
+++ b/.test/tests/templates-resolver-ipv6/expected-std-out.txt
@@ -1,0 +1,2 @@
+example.com - OK
+ipv6 nameserver(s) present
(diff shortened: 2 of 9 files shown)


thresheek:
Hi @serverwentdown !

Thanks for your contribution - the changes look reasonable to me.

May I ask you to extend [https://github.com/nginxinc/docker-nginx/blob/master/.test/config.sh](config.sh) to automatically run the tests?

thresheek:
I went ahead and added templates-resolver-ipv6 to the tests run and fixed an actual test to:

- properly register exit traps
- use ipv6 address of a test container to connect to
- check for ipv6 addresses being present in the configuration

The last check depends on the environment behaviour which isnt great, but I don't think we can easily emulate editing /etc/resolv.conf in this case, so we have to rely on GH runners to do the correct thing for us.

PR #857, opened 2024-01-08
Replace `COPY` with bind mount in modules examples

jnoordsij:
### Proposed changes

In the modules examples, a `COPY` from the builder is done to ensure it's output is available to the final image. However, this operation results in the layer being part of the final image and thus slightly bloating it. 

By using a bind mount for the `RUN` operation, the builder content is exposed to this particular run operation only, in a read-only manner, and will not be part of the final image in any way.

See also https://docs.docker.com/engine/reference/builder/#run---mount.

### Checklist

### DIFF
--- a/.test/tests/modules/run.sh
+++ b/.test/tests/modules/run.sh
@@ -32,7 +32,7 @@ fi
 
 # Create an instance of the container-under-test
 modulesImage="$("$HOME/oi/test/tests/image-name.sh" librarytest/nginx-template "$image")"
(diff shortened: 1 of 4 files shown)


thresheek:
Hi @jnoordsij!

It looks like the tests fail since we're using non-buildkit builds... This means we'd need to adapt the tests and document this on modules/ for users to be aware of the limitation.

jnoordsij:
> Hi @jnoordsij!
> 
> It looks like the tests fail since we're using non-buildkit builds... This means we'd need to adapt the tests and document this on modules/ for users to be aware of the limitation.

Ah yeah, I didn't realise this would be enforced in these builds as well. I've updated the tests and added some documentation notes. Let's hope the tests will succeed now!

PR #874, opened 2024-03-21
Keep repo key registered in alpine, and simplify Dockerfiles

orgads:
### Proposed changes

Simplify package installation in non-slim variants.

This can also be used in user images to install packages directly.

### Checklist

- [X] I have read the [`CONTRIBUTING`](https://github.com/nginxinc/docker-nginx/blob/master/CONTRIBUTING.md) document
- [X] I have run `./update.sh` and ensured all entrypoint/Dockerfile template changes have been applied to the relevant image entrypoint scripts & Dockerfiles
- [X] If applicable, I have added tests that prove my fix is effective or that my feature works
- [X] If applicable, I have checked that any relevant tests pass after adding my changes
- [X] I have updated any relevant documentation

### DIFF
--- a/Dockerfile-alpine-otel.template
+++ b/Dockerfile-alpine-otel.template
@@ -12,17 +12,7 @@ RUN set -x \
     && case "$apkArch" in \
         x86_64|aarch64) \
 # arches officially built by upstream
-            set -x \
-            && KEY_SHA512="e09fa32f0a0eab2b879ccbbc4d0e4fb9751486eedda75e35fac65802cc9faa266425edf83e261137a2f4d16281ce2c1a5f4502930fe75154723da014214f0655" \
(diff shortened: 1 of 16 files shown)


thresheek:
Hi @orgads !

We did that because we explicitely did not want the image to have the repository configured - as adding more packages with the repos configured might upgrade the packages to something newer than specified initially.  This is also consistent with what we have for Debian-based images.

orgads:
I see. I now changed it to just keep the key registered, but not register the repo. I see this is also done in debian. What do you say?

orgads:
And simplified debian perl and otel to use the existing key too.

orgads:
@thresheek I think the build fails because it tries to build all alpine in parallel, so the non-slim variants use the existing image as base, and this change introduces a dependency between slim and non-slim variants.

PR #913, opened 2024-08-01
fix: remove trailing whitespace from awk command

lockejan:
* trailing whitespaces break configs that use quotation marks around vars. See https://github.com/nginxinc/docker-nginx-unprivileged/issues/234

### Proposed changes

Currently the awk command in `entrypoint/15-local-resolvers.envsh` will always produce a trailing whitespace.
This will break configs that make use of quotation marks around variable in config templates.
Initially opened the issue here: https://github.com/nginxinc/docker-nginx-unprivileged/issues/234

### Checklist

- [ ] I have run `./update.sh` and ensured all entrypoint/Dockerfile template changes have been applied to the relevant image entrypoint scripts & Dockerfiles
- [ ] If applicable, I have checked that any relevant tests pass after adding my changes

### DIFF
--- a/entrypoint/15-local-resolvers.envsh
+++ b/entrypoint/15-local-resolvers.envsh
@@ -9,4 +9,7 @@ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 [ "${NGINX_ENTRYPOINT_LOCAL_RESOLVERS:-}" ] || return 0
 
 NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {if ($2 ~ ":") {print "["$2"]"} else {print $2}}' /etc/resolv.conf)
+
+NGINX_LOCAL_RESOLVERS="${NGINX_LOCAL_RESOLVERS% }"
+
 export NGINX_LOCAL_RESOLVERS
--- a/mainline/alpine-slim/15-local-resolvers.envsh
+++ b/mainline/alpine-slim/15-local-resolvers.envsh
@@ -9,4 +9,7 @@ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 [ "${NGINX_ENTRYPOINT_LOCAL_RESOLVERS:-}" ] || return 0
 
 NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {if ($2 ~ ":") {print "["$2"]"} else {print $2}}' /etc/resolv.conf)
+
+NGINX_LOCAL_RESOLVERS="${NGINX_LOCAL_RESOLVERS% }"
+
 export NGINX_LOCAL_RESOLVERS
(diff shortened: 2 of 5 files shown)


thresheek:
I don't think it's a proper fix since we can have multiple resolvers defined, and then if ORS is set to "" they will be squashed together, e.g. "127.0.0.1127.0.0.2".

thresheek:
We probably just want to check if the latest character of the variable is whitespace and trim it

thresheek:
I'd prefer a posix shell-compatible solution rather than another (costly) awk call - e.g. get the last character of a string, and then compare/trim

lockejan:
> I'd prefer a posix shell-compatible solution rather than another (costly) awk call - e.g. get the last character of a string, 
and then compare/trim

Totally see your point.
However, I'm not an expert to be honest.
I found at least two ways of achieving this. Personally I like A better.

A. xargs

```sh
NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {if ($2 ~ ":") {print "["$2"]"} else {print $2}}' /etc/resolv.conf | xargs)
```

B. shell expansion
```sh
NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {if ($2 ~ ":") {print "["$2"]"} else {print $2}}' /etc/resolv.conf)
NGINX_LOCAL_RESOLVERS="${NGINX_LOCAL_RESOLVERS%"${NGINX_LOCAL_RESOLVERS##*[![:space:]]}"}"
```

thresheek:
I think I've missed the last comment - sorry about that.

Wouldnt the following simple patch work, as we're explicitely setting ORS to a space?

```
diff --git a/entrypoint/15-local-resolvers.envsh b/entrypoint/15-local-resolvers.envsh
index 450a999..e830dda 100755
--- a/entrypoint/15-local-resolvers.envsh
+++ b/entrypoint/15-local-resolvers.envsh
@@ -9,4 +9,7 @@ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 [ "${NGINX_ENTRYPOINT_LOCAL_RESOLVERS:-}" ] || return 0

 NGINX_LOCAL_RESOLVERS=$(awk 'BEGIN{ORS=" "} $1=="nameserver" {if ($2 ~ ":") {print "["$2"]"} else {print $2}}' /etc/resolv.conf)
+
+NGINX_LOCAL_RESOLVERS="${NGINX_LOCAL_RESOLVERS% }"
+
 export NGINX_LOCAL_RESOLVERS
```

thresheek:
I also thought about adding a test to catch issues like that, but more I look into, more I think it's a misconfiguration to put quotation marks in there, e.g. `resolver "127.0.0.1 127.0.0.2";` results in:
```
nginx: [emerg] host not found in resolver "127.0.0.1 127.0.0.2" in /etc/nginx/conf.d/test.conf:3
```

lockejan:
No worries. Thanks for finalising it. I never used multiple resolvers yet, so I wasn't really aware of that test case. But yeah it's rather a misconfiguration and the trailing whitespace was only problematic for this edge case. Will update my config. ✌️

PR #922, opened 2024-08-17
Use packaging sources to build packages on non-mainstream architectures.

thresheek:
### Proposed changes

Similarly to what's currently happening on Alpine, use pkg-oss sources to build packages from source on Debian.  This allows us to provide fixes for architectures not covered by upstream nginx.org packages (amd64, aarch64) without touching the rest of infrastructure on nginx.org.

### Checklist

- [x] If applicable, I have added tests that prove my fix is effective or that my feature works
- [x] I have updated any relevant documentation

### DIFF
--- a/Dockerfile-debian-otel.template
+++ b/Dockerfile-debian-otel.template
@@ -15,24 +15,46 @@ RUN set -x; \
             ;; \
         *) \
 # we're on an architecture upstream doesn't officially build for
-# let's build binaries from the published source packages
-            echo "deb-src [signed-by=$NGINX_GPGKEY_PATH] %%PACKAGEREPO%% %%DEBIAN_VERSION%% nginx" >> /etc/apt/sources.list.d/nginx.list \
-            \
+# let's build binaries from the published packaging sources
 # new directory for storing sources and .deb files
-            && tempDir="$(mktemp -d)" \
+            tempDir="$(mktemp -d)" \
             && chmod 777 "$tempDir" \
 # (777 to ensure APT's "_apt" user can access it too)
             \
 # save list of currently-installed packages so build dependencies can be cleanly removed later
             && savedAptMark="$(apt-mark showmanual)" \
             \
-# build .deb files from upstream's source packages (which are verified by apt-get)
+# build .deb files from upstream's packaging sources
             && apt-get update \
-            && apt-get build-dep -y %%BUILDTARGET%% \
+            && apt-get install --no-install-recommends --no-install-suggests -y \
+                curl \
+                devscripts \
+                equivs \
+                git \
+                libxml2-utils \
+                lsb-release \
+                xsltproc \
             && ( \
                 cd "$tempDir" \
-                && DEB_BUILD_OPTIONS="nocheck parallel=$(nproc)" \
-                    apt-get source --compile %%BUILDTARGET%% \
+                && REVISION="%%REVISION%%" \
+                && REVISION=${REVISION%~*} \
+                && curl -f -O https://hg.nginx.org/pkg-oss/archive/${REVISION}.tar.gz \
+                && PKGOSSCHECKSUM="%%PKGOSSCHECKSUM%% *${REVISION}.tar.gz" \
+                && if [ "$(openssl sha512 -r ${REVISION}.tar.gz)" = "$PKGOSSCHECKSUM" ]; then \
+                    echo "pkg-oss tarball checksum verification succeeded!"; \
+                else \
+                    echo "pkg-oss tarball checksum verification failed!"; \
+                    exit 1; \
+                fi \
+                && tar xzvf ${REVISION}.tar.gz \
+                && cd pkg-oss-${REVISION} \
+                && cd debian \
+                && for target in %%BUILDTARGET%%; do \
+                    make rules-$target; \
+                    mk-build-deps --install --tool="apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes" \
+                        debuild-$target/nginx-$NGINX_VERSION/debian/control; \
+                done \
+                && make %%BUILDTARGET%% \
             ) \
 # we don't remove APT lists here because they get re-downloaded and removed later
             \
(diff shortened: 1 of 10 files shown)


Dockerfile-debian-otel.template
```diff
@@ -15,24 +15,57 @@ RUN set -x; \
...
+                libedit-dev \
+                libxml2-dev \
+                libxslt-dev \
+                cmake \
+                libre2-dev \
+                libc-ares-dev \
+                pkg-config \
+                libperl-dev \
+                libxml2-utils \
+                xsltproc \
```
defanator, line 52: Would be cool to get these sorted.

Dockerfile-debian-perl.template
```diff
@@ -13,24 +13,57 @@ RUN set -x; \
...
+                libedit-dev \
+                libxml2-dev \
+                libxslt-dev \
+                cmake \
+                libre2-dev \
+                libc-ares-dev \
+                pkg-config \
+                libperl-dev \
+                libxml2-utils \
+                xsltproc \
```
defanator, line 50: Ditto (re: sort the list).

Dockerfile-debian.template
```diff
@@ -43,24 +43,57 @@ RUN set -x \
...
+                libedit-dev \
+                libxml2-dev \
+                libxslt-dev \
+                cmake \
+                libre2-dev \
+                libc-ares-dev \
+                pkg-config \
+                libperl-dev \
+                libxml2-utils \
+                xsltproc \
```
defanator, line 80: Ditto (re: sort the list).

mainline/debian-otel/Dockerfile
```diff
@@ -26,24 +26,57 @@ RUN set -x; \
...
+                libedit-dev \
+                libxml2-dev \
+                libxslt-dev \
+                cmake \
+                libre2-dev \
+                libc-ares-dev \
+                pkg-config \
+                libperl-dev \
+                libxml2-utils \
+                xsltproc \
```
defanator, line 63: Ditto (re: sort the list).

mainline/debian-perl/Dockerfile
```diff
@@ -24,24 +24,57 @@ RUN set -x; \
...
+                libedit-dev \
+                libxml2-dev \
+                libxslt-dev \
+                cmake \
+                libre2-dev \
+                libc-ares-dev \
+                pkg-config \
+                libperl-dev \
+                libxml2-utils \
+                xsltproc \
```
defanator, line 61: Ditto (re: sort the list).

mainline/debian/Dockerfile
```diff
@@ -53,24 +53,57 @@ RUN set -x \
...
+                libedit-dev \
+                libxml2-dev \
+                libxslt-dev \
+                cmake \
+                libre2-dev \
+                libc-ares-dev \
+                pkg-config \
+                libperl-dev \
+                libxml2-utils \
+                xsltproc \
```
defanator, line 90: Ditto (re: sort the list).

stable/debian-otel/Dockerfile
```diff
@@ -26,24 +26,57 @@ RUN set -x; \
...
+                libedit-dev \
+                libxml2-dev \
+                libxslt-dev \
+                cmake \
+                libre2-dev \
+                libc-ares-dev \
+                pkg-config \
+                libperl-dev \
+                libxml2-utils \
+                xsltproc \
```
defanator, line 63: Ditto (re: sort the list).

stable/debian-perl/Dockerfile
```diff
@@ -24,24 +24,57 @@ RUN set -x; \
...
+                libedit-dev \
+                libxml2-dev \
+                libxslt-dev \
+                cmake \
+                libre2-dev \
+                libc-ares-dev \
+                pkg-config \
+                libperl-dev \
+                libxml2-utils \
+                xsltproc \
```
defanator, line 61: Ditto (re: sort the list).

stable/debian/Dockerfile
```diff
@@ -53,24 +53,57 @@ RUN set -x \
...
+                libedit-dev \
+                libxml2-dev \
+                libxslt-dev \
+                cmake \
+                libre2-dev \
+                libc-ares-dev \
+                pkg-config \
+                libperl-dev \
+                libxml2-utils \
+                xsltproc \
```
defanator, line 90: Ditto (re: sort the list).

defanator (approved):
This is nice! Minor comments inside.

Dockerfile-debian-otel.template
thresheek: Yeah, I initially put the list directly from the packages build-depends; but it probably makes sense to make to sort alphabetically here.

tianon: Maybe there's a clever way we can emulate what `sbuild` and friends do to get these installed directly via `Build-Depends` itself so you don't have to maintain the list in two places? :thinking:

tianon: Ah, `mk-build-deps` from `devscripts` should do the trick, if not copying this full dependency list is something that interests you :eyes:

tianon: (It's also possible something like `apt-get build-dep .` would work, but I don't know whether that's only available in newer versions of APT than we've got in the stable releases.)

thresheek: Yeah, I'd rather not maintain it here as well;  I'll take a look if we can have something like `mk-build-deps` do the work here.

Maybe also something like https://github.com/nginx/unit/blob/master/pkg/deb/Makefile#L133 could be implemented in `pkg-oss`.

tianon: Ah, in that case, something like `dpkg-checkbuilddeps` might be interesting too :eyes:

thresheek: erf, I think mk-build-deps is actually a much nicer way, and it turns out (lol) I've already implemented it for modules in https://github.com/nginxinc/docker-nginx/blob/5b7624a8f98af507d6d1da2ad7b01806305524ae/modules/Dockerfile#L61 🙄 🤣

thresheek: Anyway, I've pushed a patch that fixes this and uses mk-build-deps.

PR #974, opened 2025-03-21
modules/README.md: updated the current list of modules.

thresheek:
Updates the current list of modules.


thresheek:
I have hereby read the F5 CLA and agree to its terms

thresheek:
@alessfg the CLA check seems to fail still.  Is that because this repo has the master branch instead of a main?

thresheek:
recheck

alessfg:
Fixed the issue in #975 :)

PR #1007, opened 2025-10-28
acme module support for mainline and stable images

thresheek:
### Proposed changes

This PR adds support for nginx-module-acme install in the base image.

This requires `stable` images to be bumped to newer Operating Systems, and a few changes to the build system, notably installing cargo, and setting CARGO_HOME properly.


thresheek:
Now that mainline and stable are bumped to the contemporary distros, we don't have any blockers for the acme support.

Please review.

PR #1016, opened 2026-01-12
Drop arm32v6 for Alpine-derived images

thresheek:
### Proposed changes

DOI infra seems to struggle with building Rust-based acme module for
this specific target.

Given that Debian builds for the same images are built just fine, users
of said devices (Raspberry Pi 0 and Raspberry Pi 1) have a
migration path to the newer NGINX versions.

An alternative is to introduce additional branching in
update.sh/Dockerfile templates, but that seems excessive just to support
this particular target with a limited number of real world users.

cc @yosifkit @tianon

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -21,8 +21,8 @@ debian_architectures=(
 
 declare -A alpine_architectures
 alpine_architectures=(
-	[mainline]='arm64v8, arm32v6, arm32v7, ppc64le, s390x, i386, amd64, riscv64'
-	[stable]='arm64v8, arm32v6, arm32v7, ppc64le, s390x, i386, amd64, riscv64'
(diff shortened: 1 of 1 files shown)


thresheek:
To add a bit of context, it seems DOI builds are stuck somewhere during the Rust build.  I was not able to reproduce the issue on a local instance of podman (4 CPU, 4G RAM) running on top of M4 Mac under the current FC43-based coreOS VM.  

Not sure what other options we have and I would rather not abuse the build infra with periodic rebuilds of things that cant be rebuilt; adding per-architecture branches to `update.sh` also seems to be a bit excessive.

LMK what you all think. Thanks

thresheek:
I believe https://github.com/nginx/docker-nginx/pull/1017 offers a better outcome for our users.  Closing this PR.

PR #1043, opened 2026-05-13
Bump nginx mainline and stable

thresheek:
### Proposed changes

This PR bumps nginx mainline and stable to current versions.

Also, njs is updated as well.


oxpa (commented):
LGTM

PR #1058, opened 2026-07-15
Bump versions for nginx, njs, and alpine

thresheek:
### Proposed changes

This PR:
- bumps nginx to current for stable and mainline
- bumps njs to 1.0.0 (thanks @orgads in https://github.com/nginx/docker-nginx/pull/1056)
- bumps Alpine Linux to 3.24 (thanks @orgads in https://github.com/nginx/docker-nginx/pull/1052)

### DIFF
--- a/.test/tests/modules/nginx.conf.sme
+++ b/.test/tests/modules/nginx.conf.sme
@@ -3,7 +3,6 @@ worker_processes  auto;
 
 load_module modules/ndk_http_module.so;
 load_module modules/ngx_http_echo_module.so;
-load_module modules/ngx_http_set_misc_module.so;
 
 error_log  /var/log/nginx/error.log notice;
 pid        /var/run/nginx.pid;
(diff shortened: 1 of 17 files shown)


eepifanova (approved):

thresheek:
Looking on why it's all red...

mschfh:
@thresheek Any thoughts about adding module (smoke) tests to `pkg-oss`?

oxpa (approved):

thresheek:
hey @mschfh,

Indeed, `pkg-oss` tests whether the subset of first-party and third-party modules do build with the current change, but we don't specifically test whether they actually work.  I'm not even sure it's a good idea to do that for third-party modules (and block the PR acceptance) since it's OK for NGINX to introduce breaking changes, and module authors are supposed to follow up.

mschfh:
@thresheek Makes sense, what about non-blocking tests?
From my understanding, the current (contrib) code in the `pkg-oss` repo is broken without any indication of it.
Additional tests would also help to prevent regressions (or validate fixes) when updating the module version itself.

thresheek:
@mschfh I fully agree.  Let's move the discussion over to https://github.com/nginx/pkg-oss/issues/125 - much appreciated!

